PR #20, opened 2012-10-17
Function `execurl` in preloads.py to load an external Python file from an URL

versae:
Added the function `execurl` to load an external Python file from an URL and `exec` it as if were local. This addition allows you to have your convenient functions library stored in GitHub, gist.github.com, share them through Dropbox, etc.

Also it opens a door to load external graphs from URLs.


versae:
After 4 years, any reason for the closing? IDK why all those commits are there, my original PR was like 3 lines of code.

eduramiba:
Hi,
I think this was automatically closed by github when I force-pushed the new code for scripting plugin, which now uses the new build system for plugins.
I wasn't aware of this pull request but I will check your original code to see if it can be included in the release of scripting plugin for 0.9.1

Thanks,
Eduardo

versae:
Oh, I see. Thanks by the way.

eduramiba:
Done https://github.com/gephi/gephi-plugins/commit/fbffdb6156d39f38551c318ef219266fe9714ce4
Thanks!

PR #69, opened 2015-12-27
JsonExporter Plugin

computermacgyver:
A basic plugin to export a graph to a file in JSON format. This is an update of the plugin for version 0.8.2.

Awesome work on the new version and the overhaul of the plugin system. It looks like master-forge is behind master; so, the pull request captures some differences between master and master-forge.

Thanks!

### DIFF
--- a/modules/JsonExporter/pom.xml
+++ b/modules/JsonExporter/pom.xml
@@ -1,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>gephi-plugin-parent</artifactId>
+        <groupId>org.gephi</groupId>
+        <version>0.9.0</version>
+    </parent>
+
+    <groupId>uk.ox.oii</groupId>
+    <artifactId>jsonexporter-plugin</artifactId>
+    <version>2.0.0</version>
+    <packaging>nbm</packaging>
+
+    <name>Json Exporter</name>
+
+    <dependencies>
+        <!-- Insert dependencies here -->
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>nbm-maven-plugin</artifactId>
+                <configuration>
+                    <licenseName>Apache 2.0</licenseName>
+                    <!--<licenseFile></licenseFile>-->
+                    <author>Scott A. Hale</author>
+                    <authorEmail></authorEmail>
+                    <authorUrl>http://www.scotthale.net/</authorUrl>
+                    <sourceCodeUrl>http://github.com/oxfordinternetinstitute/gephi-plugins</sourceCodeUrl>
+                    <homePageUrl></homePageUrl>
+                    <publicPackages>
+                        <!-- Insert public packages -->
+                    </publicPackages>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+</project>
+
+
--- a/modules/JsonExporter/pom.xml
+++ b/modules/JsonExporter/pom.xml
@@ -7,45 +7,45 @@
         <version>0.9.0</version>
     </parent>
 
-    <groupId>org.gephi</groupId>
+    <groupId>uk.ac.ox.oii</groupId>
     <artifactId>jsonexporter-plugin</artifactId>
     <version>2.0.0</version>
     <packaging>nbm</packaging>
-
+    
     <name>Json Exporter</name>
 
     <dependencies>
         <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
-            <version>2.2.2</version>
+            <version>2.5</version>
         </dependency>
         <dependency>
-            <groupId>${project.groupId}</groupId>
+            <groupId>org.gephi</groupId>
             <artifactId>utils-longtask</artifactId>
         </dependency>
         <dependency>
-            <groupId>${project.parent.groupId}</groupId>
+            <groupId>org.gephi</groupId>
             <artifactId>layout-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>${project.parent.groupId}</groupId>
+            <groupId>org.gephi</groupId>
             <artifactId>io-exporter-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>${project.parent.groupId}</groupId>
+            <groupId>org.gephi</groupId>
             <artifactId>preview-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>${project.parent.groupId}</groupId>
+            <groupId>org.gephi</groupId>
             <artifactId>project-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>${project.parent.groupId}</groupId>
+            <groupId>org.gephi</groupId>
             <artifactId>graph-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>${project.parent.groupId}</groupId>
+            <groupId>org.gephi</groupId>
             <artifactId>ui-propertyeditor</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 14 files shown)


modules/JsonExporter/pom.xml
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>gephi-plugin-parent</artifactId>
+        <groupId>org.gephi</groupId>
+        <version>0.9.0</version>
+    </parent>
+
+    <groupId>org.gephi</groupId>
```
mbastian: Please pick something else than org.gephi here as we reserve this for official gephi-team work at the moment

modules/JsonExporter/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+    <version>2.0.0</version>
+    <packaging>nbm</packaging>
+
+    <name>Json Exporter</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>2.2.2</version>
```
mbastian: Upgrade to 2.4, just a suggestion

modules/JsonExporter/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
```
mbastian: Update to "org.gephi". This was an error on my end to suggest this. It shouldn't be any Maven variable as it gets overriden by the project.

modules/JsonExporter/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+            <artifactId>utils-longtask</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${project.parent.groupId}</groupId>
```
mbastian: Same here, please update all Gephi dependencies groupId to org.gephi

modules/JsonExporter/src/main/java/uk/ac/ox/oii/jsonexporter/JSONExporter.java
```diff
@@ -0,0 +1,277 @@
...
+                gson.toJson(json, writer);
+
+                //Finish progress
+                Progress.finish(progress);
+                return true;
+            } else {
+                throw new FileNotFoundException("Writer is null");
+            }
+        } catch (Exception e) {
+            e.printStackTrace();
```
mbastian: This is a cleaner way to propagate Exception (i.e. the user will see it)

Logger.getLogger(JSONExporter.class.getName()).log(Level.SEVERE, null, e);

You can remove the throw new RuntimeException() then

modules/JsonExporter/src/main/java/uk/ac/ox/oii/jsonexporter/JSONExporter.java
```diff
@@ -0,0 +1,277 @@
...
+
+                HashMap<String, HashSet<GraphElement>> json = new HashMap<String, HashSet<GraphElement>>();
+                json.put("nodes", jNodes);
+                json.put("edges", jEdges);
+
+                gson.toJson(json, writer);
+
+                //Finish progress
+                Progress.finish(progress);
+                return true;
```
mbastian: This is way to keep the task consistent with cancelation:

return !cancel;

modules/JsonExporter/src/main/java/uk/ac/ox/oii/jsonexporter/JSONExporter.java
```diff
@@ -0,0 +1,277 @@
...
+        try {
+            if (writer != null) {//path.getParentFile().exists()
+                GraphModel graphModel = workspace.getLookup().lookup(GraphModel.class);
+                Graph graph = null;
+                if (exportVisible) {
+                    graph = graphModel.getGraphVisible();
+                } else {
+                    graph = graphModel.getGraph();
+                }
+
```
mbastian: This exporter isn't resistant to a writer thread modifying the graph. The exporter graph might be inconsistent if another thread delete things between the nodes and edges export for instance. I suggest to add here the line `graph.readLock();` and then in the `finally {` block add `graph.readUnlock();`.

modules/JsonExporter/src/main/java/uk/ac/ox/oii/jsonexporter/JSONExporter.java
```diff
@@ -0,0 +1,277 @@
...
+                HashSet<GraphElement> jNodes = new HashSet<GraphElement>();
+                Node[] nodeArray = graph.getNodes().toArray();
+                
+                for (Node n : nodeArray) {
+
+                    String id = n.getId().toString();
+                    String label = n.getLabel();
+                    float x = n.x();
+                    float y = n.y();
+                    System.out.println(n.z());
```
mbastian: Please remove System.out here

modules/JsonExporter/src/main/java/uk/ac/ox/oii/jsonexporter/JSONExporter.java
```diff
@@ -0,0 +1,277 @@
...
+                    GraphNode jNode = new GraphNode(id);
+                    jNode.setLabel(label);
+                    jNode.setX(x);
+                    jNode.setY(y);
+                    jNode.setSize(size);
+                    jNode.setColor(color);
+
+                    
+                    for (Column col : attModel) {
+                            String name = col.getTitle();
```
mbastian: Unlike the column's id, the column's title might be localized so it isn't recommended to do comparaison based on that. I would suggest to rather use `col.getId()`. The default ID column's id is `id` and same for `label`.

modules/JsonExporter/src/main/java/uk/ac/ox/oii/jsonexporter/JSONExporter.java
```diff
@@ -0,0 +1,277 @@
...
+
+                    boolean mixColors=false;
+                    String color="";
+                    jEdge.setLabel(e.getLabel());
+
+                    float r=e.r();
+                    float g=e.g();
+                    float b=e.b();
+
+                    if (r==-1 || g==-1 || b==-1) {
```
mbastian: The way we mark edges NOT having a color has changed and you should rather test whether alpha == 0

modules/JsonExporter/src/main/java/uk/ac/ox/oii/jsonexporter/JSONExporter.java
```diff
@@ -0,0 +1,277 @@
...
+                        //Mix colors
+                        mixColors=true;
+                    } else {
+                        color = "rgb(" + (int) (r* 255) + "," + (int) (g* 255) + "," + (int) (b* 255) + ")";
+                    }
+
+                    Iterator<Column> eAttr = e.getAttributeColumns().iterator();
+                    while (eAttr.hasNext()) {
+                        Column col = eAttr.next();
+                        if (col == null) {
```
mbastian: Null can't happen here

modules/JsonExporter/src/main/java/uk/ac/ox/oii/jsonexporter/JSONExporter.java
```diff
@@ -0,0 +1,277 @@
...
+                            continue;
+                        }
+                        String name = col.getTitle();
+                        if (name.equalsIgnoreCase("Id") || name.equalsIgnoreCase("Label")
```
mbastian: Another way to exclude all property columns is to do `!col.isProperty()`. That would exclude id, label, weight (for edge)

modules/JsonExporter/pom.xml
computermacgyver: Thank you @mbastian for this note and the others. I will make all the changes and submit an updated pull request. I thought org.gephi would be reserved, but had left it in because of the {project.groupId} reference in the dependencies. I will remove that and change this groupId. Thank you for your time to review the code so thoroughly.

computermacgyver:
Thank you, @mbastian , for the extremely helpful comments/fixes. I've made all the changes now and they have been added to this pull request.

mbastian:
Thanks @computermacgyver for your work! The plugin is now available on Gephi!

computermacgyver:
Thank you, @mbastian . I really appreciate all your help to make this happen -- and of course the new version of Gephi!

PR #72, opened 2016-01-07
Multimode network

totetmatt:


eduramiba:
Hi,
Thanks for the pull request. I will comment some details to fix in the code before we can merge it.

eduramiba:
Thanks for the fixes! I will continue reviewing it later

modules/Multimode-Networks/src/main/java/cz/cvut/fit/gephi/multimode/MultiModeWindowTopComponent.java
```diff
@@ -0,0 +1,344 @@
...
+    private void firstMatrixActionPerformed(java.awt.event.ActionEvent evt) {//GEN-FIRST:event_firstMatrixActionPerformed
+        // TODO reaction on selected first matrix                
+        secondMatrix.removeAllItems();
+        if (attributes.getSelectedIndex() > 0) {
+
+            Column col = (Column) attributes.getSelectedItem();
+
+            // load all possible values of attribute
+            GraphController graphController = Lookup.getDefault().lookup(GraphController.class);
+            Graph graph = graphController.getGraphModel().getGraphVisible();
```
eduramiba: Ideally one should check if there is any open project before accessing other controllers data such as GraphController.getGraphModel (here and in every place).

But the reality is many plugins don't do this so they will cause exceptions if no project is open.
So, if you can improve this, it would be great but we won't impose it in order to do the merge since it's not a critical problem.

This is done via ProjectController.
For example: Lookup.getDefault().lookup(ProjectController.class).getCurrentProject()
But in most cases it's useful to add a WorkspaceListener to keep track of projects opening and closing.

Real example at https://github.com/gephi/gephi/blob/master/modules/DesktopDataLaboratory/src/main/java/org/gephi/desktop/datalab/DataTableTopComponent.java#L310 from lines 310 to 347

modules/Multimode-Networks/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+      <groupId>org.netbeans.api</groupId>
+      <artifactId>org-openide-util</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.gephi</groupId>
+      <artifactId>utils-longtask</artifactId>
+    </dependency>
+        <dependency>
+            <groupId>org.gephi</groupId>
+            <artifactId>datalab-api</artifactId>
```
eduramiba: It seems this dependency is not used at all in the code so you can remove it (and remove unused imports).

modules/Multimode-Networks/src/main/java/cz/cvut/fit/gephi/multimode/GraphColoring.java
```diff
@@ -0,0 +1,123 @@
...
+        Column colorColumn;
+        
+        if (!nodeTable.hasColumn(colorColumnName)) {
+            colorColumn = nodeTable.addColumn(
+                    colorColumnName,
+                    "Node Color Multimode",
+                    String.class,
+                    Origin.DATA,
+                    NodeColor.BLACK.getValue(),
+                    false);
```
eduramiba: Indexed (false) should be true here. In Gephi we create all columns with indexed = true so they can be used in filters and appeareance.

The non-indexed option is for other possible usages of graphstore API outside of Gephi.

modules/Multimode-Networks/src/main/java/cz/cvut/fit/gephi/multimode/LongTaskTransformation.java
```diff
@@ -0,0 +1,193 @@
...
+        
+        for (int i = 0; i < result.getM(); i++) {
+            for (int j = 0; j < result.getN(); j++) {
+                if (graph.contains(firstVertical.get(i)) && graph.contains(secondHorizontal.get(j)) && graph.getEdge(firstVertical.get(i), secondHorizontal.get(j)) == null && result.get(i, j) > 0) {
+                    Edge ee = graphModel.factory().newEdge(firstVertical.get(i), secondHorizontal.get(j),1, (float) result.get(i, j), graphModel.isDirected());
+                    if (!ee.isSelfLoop()) {
+                        
+                        ee.setAttribute(MMNT, inDimension + "<--->" + outDimension);
+                        ee.setLabel(inDimension + "-" + outDimension);
+                        graph.addEdge(ee);
```
eduramiba: This part of the code should check if the same edge already exists before adding it.

I uploaded sample code for this and other changes here: https://github.com/eduramiba/gephi-plugins/tree/multimode-network

modules/Multimode-Networks/src/main/java/cz/cvut/fit/gephi/multimode/LongTaskTransformation.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+        Progress.progress(progressTicket, "Removing nodes/edges");
+        
+        
+        if (removeNodes) {
+            for (Node n : firstHorizontal) {
+                graph.removeNode(n);
+            }
+        } else {
+            if (removeEdges) {
```
eduramiba: No changes required here. Just note that this can cause an exception due to this issue https://github.com/gephi/graphstore/issues/103 that will be fixed in a future update.

eduramiba:
Thank you,
I just noticed you made the changes, I kind of expected that github would alert me!

PR #74, opened 2016-01-19
Sigmaexporter plugin

computermacgyver:
This is an update of the existing plugin for version 0.8.2. This first port of the plugin provides a quick solution for those looking for the same functionality available in 0.8.2. The HTML/JavaScript has not yet been updated to newer versions, but I aim do so in the near future.


eduramiba:
Hi Scott,
Thanks for the pull request :)

I will post a few comments for improvement before we do the pull request.

modules/sigmaExporter/src/main/java/uk/ac/ox/oii/sigmaexporter/SigmaExporterUI.java
```diff
@@ -0,0 +1,147 @@
...
+ * control the rest.
+ * <p>
+ * The <code>action()</code> creates a new exporter and configure it with
+ * the settings panel. The export process is wrapped in a {@link LongTaskExecutor}
+ * which executes {@link LongTask} in a separated thread and supports progress
+ * and cancel.
+ * 
+ * 
+ * @author Mathieu Bastian
+ */
```
eduramiba: Incorrect copy pasted javadoc here, you can just remove it.

Same with license. Also, some java files don't have any license (they have the netbeans default "To change this template, choose Tools | Templates"). Maybe you can set the same one for all the java files.

modules/sigmaExporter/src/main/java/uk/ac/ox/oii/sigmaexporter/ZipHandler.java
```diff
@@ -0,0 +1,135 @@
...
+    }*/
+
+    public static void extractZip(InputStream input, String dest) throws Exception {
+        byte[] buf = new byte[1024];
+        ZipInputStream zinstream = new ZipInputStream(input);
+        ZipEntry zentry = zinstream.getNextEntry();
+        System.out.println("Name of current Zip Entry : " + zentry + "\n");
+        while (zentry != null) {
+            String entryName = zentry.getName();
+            System.out.println("Name of  Zip Entry : " + entryName);
```
eduramiba: We are trying to not print anything to System.out. You can use ```javaLogger.getLogger("").info("message");``` for example

modules/sigmaExporter/src/main/java/uk/ac/ox/oii/sigmaexporter/ZipHandler.java
```diff
@@ -0,0 +1,135 @@
...
+
+import java.io.*;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.SimpleFileVisitor;
+import java.nio.file.StandardCopyOption;
```
eduramiba: Many unnecessary imports and commented code here. At least I would remove unused imports in this and other files.

modules/sigmaExporter/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>nbm-maven-plugin</artifactId>
+                <configuration>
+                    <licenseName>GPL v3</licenseName>
+                    <licenseFile>gpl-3.0.txt</licenseFile>
+                    <author>Scott A. Hale, Oxford Internet Institute</author>
```
eduramiba: You could set the same authorUrl as in https://github.com/oxfordinternetinstitute/gephi-plugins/blob/jsonexporter-plugin/modules/JsonExporter/pom.xml

computermacgyver:
Thanks, @eduramiba for the helpful comments. I didn't look as in-depth at the support files that I didn't modify: apologies. I've made all the changes and updated this pull request. I think it is now ready to be merged, but I am very happy to make further changes if necessary. Many thanks!

eduramiba:
Thanks Scott, this is now merged.
I changed the module folder name to SigmaExporter instead of sigmaExporter to keep name conventions and fixed the logger usage.

computermacgyver:
Thank you, @eduramiba . This is wonderful. Thank you for the help on this and for all the great work on Gephi.

PR #75, opened 2016-01-19
Update Graph Streaming Plugin to 0.9.0

panisson:
This is an update of the existing plugin for version 0.8.2. This first port of the plugin provides a quick solution for those looking for the same functionality available in 0.8.2.

The following functionalities have not yet been updated, but they will be updated once the proper API is available:
- Generate streaming events when attribute values are changed.


build.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mbastian: This file isn't necessary and can be deleted. It's a legacy file from the Ant-based plugin.

modules/DesktopStreaming/pom.xml
```diff
@@ -0,0 +1,104 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>gephi-plugin-parent</artifactId>
+        <groupId>org.gephi</groupId>
+        <version>0.9.0</version>
+    </parent>
+
+    <groupId>org.gephi.desktop</groupId>
```
mbastian: You should pick a groupId different from "org.gephi" as we want to reserve this to official Gephi plugins. I would also suggest to revise your naming strategy so your groupId is the same for all of the plugin's modules, and then vary the artifactId, for instance 'streaming-plugin', 'streaming-api', 'desktop-streaming' etc.

modules/DesktopStreaming/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.gephi</groupId>
+      <artifactId>lib.validation</artifactId>
+    </dependency>
+    
+    <dependency> 
+        <groupId>org.gephi.streaming</groupId>
+    <artifactId>api</artifactId>
+    <version>1.0.0</version>
```
mbastian: When you gonna update this plugin later to say, version 1.0.1 you would need to be careful and update this dependencies as well. A trick to avoid this issue is to use `${project.version}` instead here so it would the version defined above in `<version>`. That only counts if you always plan to update all of your module version at the same time . If not, you can keep that like this.

modules/GraphStreaming/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>nbm-maven-plugin</artifactId>
+                <configuration>
+                    <author>Andre Panisson</author>
+                    <sourceCodeUrl>$sourcecode_url</sourceCodeUrl>
```
mbastian: Please set source code url, for instance https://github.com/panisson/gephi-plugins/tree/graph-streaming

modules/GraphStreaming/src/main/resources/org/gephi/graphstreaming/Bundle.properties
```diff
@@ -0,0 +1,8 @@
+OpenIDE-Module-Display-Category=Import
+OpenIDE-Module-Long-Description=\
```
mbastian: This formatting isn't proper and we'll have issue displaying it on the website. Can you please put everything on one line? You can actually use html if you want, example here: https://github.com/mbastian/gephi-plugins/blob/kbrace-filter/modules/KBraceFilter/src/main/resources/org/gephi/plugins/filter/kbrace/Bundle.properties

modules/DesktopStreaming/src/main/java/org/gephi/desktop/streaming/StreamingUIController.java
```diff
@@ -0,0 +1,396 @@
...
+     *
+     * @param connection
+     */
+    public void synchronize(StreamingConnection connection) {
+
+        // Get active graph instance - Project and Graph API
+        ProjectController projectController = Lookup.getDefault().lookup(ProjectController.class);
+        Project project = projectController.getCurrentProject();
+        if (project==null)
+            projectController.newProject();
```
mbastian: Can you clarify if that's actually possible? I'm not sure in which scenario you would have to create the project. I'm asking because that isn't the right way to create a project when their a GUI (it is if done from toolkit for instance) and one should rather use the `ProjectControllerUI.newProject()` in `desktop-project` module instead.

modules/DesktopStreaming/src/main/java/org/gephi/desktop/streaming/StreamingUIController.java
```diff
@@ -0,0 +1,396 @@
...
+        Workspace workspace = projectController.getCurrentWorkspace();
+        if (workspace==null)
+            workspace = projectController.newWorkspace(projectController.getCurrentProject());
+//        projectController.openWorkspace(workspace);
+
+        GraphController graphController = Lookup.getDefault().lookup(GraphController.class);
+        GraphModel graphModel = graphController.getGraphModel();
```
mbastian: It's okay but ideally you would use the `workspace` you have and use `getGraphModel(workspace)`

modules/StreamingAPI/src/main/java/org/gephi/streaming/api/Graph2EventListener.java
```diff
@@ -0,0 +1,253 @@
...
+//            }
+//
+//            break;
+//        }
+    }
+
+    private Map<String, Object> getNodeAttributes(Node node) {
+        Map<String, Object> attributes = new HashMap<String, Object>();
+        for (Column column: node.getAttributeColumns()) {
+            if (column.getIndex() != GraphStoreConfiguration.ELEMENT_ID_INDEX) {
```
mbastian: I would suggest to replace this by !column.isProperty(). That would ignore the `id` but also the `label` and the `timeset` if defined.

modules/StreamingAPI/src/main/java/org/gephi/streaming/api/Graph2EventListener.java
```diff
@@ -0,0 +1,253 @@
...
+            attributes.put("size", node.size());
+        }
+
+        return attributes;
+    }
+
+    private Map<String, Object> getEdgeAttributes(Edge edge) {
+        Map<String, Object> attributes = new HashMap<String, Object>();
+        for (Column column: edge.getAttributeColumns()) {
+            if (column.getIndex() != GraphStoreConfiguration.ELEMENT_ID_INDEX) {
```
mbastian: Same here, with `!column.isProperty()` you would ignore the id, the label, the weight and the timeset if defined

modules/StreamingAPI/src/main/java/org/gephi/streaming/api/GraphUpdaterEventHandler.java
```diff
@@ -0,0 +1,563 @@
...
+            
+            edge = factory.newEdge(edgeId, source, target,0 , 1.0f, directed);
+            
+            if (attributes!=null && attributes.size() > 0) {
+                for(Map.Entry<String, Object> entry: attributes.entrySet()) {
+                    this.addEdgeAttribute(edge, entry.getKey(), entry.getValue());
+                }
+            }
+            
+            graph.writeLock();
```
mbastian: You don't need to wrap a `addEdge` call with a writeLock as it's done for you under the hood. However it could make sense to extend the write lock to the entire block. I can see an issue if the source or target node gets removed between the time you query them and the time you create and add the edge.

modules/StreamingAPI/src/main/java/org/gephi/streaming/api/GraphUpdaterEventHandler.java
```diff
@@ -0,0 +1,563 @@
...
+            graph.writeUnlock();
+        } else {
+            log("Edge changed event ignored for edge "+edgeId+": Edge not found");
+        }
+    }
+
+    private void edgeRemoved(String edgeId) {
+        Edge edge = graph.getEdge(edgeId);
+        if (edge!=null) {
+            graph.writeLock();
```
mbastian: Same here, remove the write lock as it's done for you behind the scene.

modules/StreamingAPI/src/main/java/org/gephi/streaming/api/GraphUpdaterEventHandler.java
```diff
@@ -0,0 +1,563 @@
...
+        if (node==null) {
+            node = factory.newNode(nodeId);
+            
+            if (attributes!=null && attributes.size() > 0) {
+                for(Map.Entry<String, Object> entry: attributes.entrySet()) {
+                    this.addNodeAttribute(node, entry.getKey(), entry.getValue());
+                }
+            }
+            
+            graph.writeLock();
```
mbastian: Remove write lock

modules/StreamingAPI/src/main/java/org/gephi/streaming/api/GraphUpdaterEventHandler.java
```diff
@@ -0,0 +1,563 @@
...
+            graph.writeUnlock();
+        } else {
+            log("Node changed event ignored for node "+nodeId+": Node not found");
+        }
+    }
+
+    private void nodeRemoved(String nodeId) {
+        Node node = graph.getNode(nodeId);
+        if (node!=null) {
+            graph.writeLock();
```
mbastian: remove write lock

modules/StreamingAPI/src/main/java/org/gephi/streaming/api/GraphUpdaterEventHandler.java
```diff
@@ -0,0 +1,563 @@
...
+
+            for (Edge edge: graph.getEdges(node).toArray()) {
```
mbastian: You don't need to do that. Removing a node will take care of removing edges. I'll update the documentation to make that clear.

modules/StreamingAPI/src/main/java/org/gephi/streaming/api/GraphUpdaterEventHandler.java
```diff
@@ -0,0 +1,563 @@
...
+        }
+    }
+    
+    private void addNodeAttribute(Node node, String attributeName, Object value) {
+        NodeProperties p = properties.getNodeProperty(attributeName);
+        if (p != null) {
+            injectNodeProperty(p, value, node);
+        }
+        
+        else if (!node.getAttributeKeys().isEmpty()) {
```
mbastian: Not sure what this line is for?

modules/StreamingAPI/src/main/java/org/gephi/streaming/api/GraphUpdaterEventHandler.java
```diff
@@ -0,0 +1,563 @@
...
+            Map<String, Object> attributes = filterEvent.getAttributes();
+            for (Node node: graph.getNodes().toArray()) {
+                if (nodeFilter.evaluate(graph, node)) {
+
+                    switch (filterEvent.getEventType()) {
+                        case ADD:
+                            log("Unsupported FilterEvent of type ADD");
+                            break;
+                        case CHANGE:
+                            graph.writeLock();
```
mbastian: Would probably be better to wrap the entire method on the writeLock to avoid side effects

mbastian:
Some additional comments as the diff is a little long for GitHub:

What is this file? `modules/StreamingImpl/src/main/resources/org/gephi/streaming/impl/cacerts`

The file `modules/StreamingImpl/src/main/resources/org/gephi/streaming/impl/layer.xml` can probably be deleted, it’s empty.

Looks like the `modules/StreamingServer/pom.xml` repeats the Jetty dependencies, can this be replaced by a dependency on `JettyWrapper`?

 In `modules/StreamingServer/src/main/java/org/gephi/streaming/server/impl/GraphWriter.java`, maybe use `column.getId()` instead of `getTitle()`?

What is this file? `modules/StreamingServer/src/main/resources/org/gephi/streaming/server/impl/localhost.p12`

Also delete `modules/StreamingServer/src/main/resources/org/gephi/streaming/server/layer.xml`?

jeromatron:
Just curious - is this likely to make it in soon?  It's a very useful plugin, but it would be great to have it work with the current version.

panisson:
@mbastian: I did try to address most of the comments.

About the file `modules/StreamingServer/src/main/resources/org/gephi/streaming/server/impl/localhost.p12`, this is an encryption key for ssl in the StreamingServer. It's not the perfect way for security, but it should just add support for https in the StreamingServer. Anyone worried about security in this case should point to his own encryption key.

modules/DesktopStreaming/src/main/java/org/gephi/desktop/streaming/StreamingUIController.java
panisson: I already addressed this, not sure why the comment is not showing up as "outdated"...

mbastian:
Cool! Looks good to me. Last request, can you merge from master and update your module poms to 0.9.1:
```
<parent>
        <artifactId>gephi-plugin-parent</artifactId>
        <groupId>org.gephi</groupId>
        <version>0.9.1</version>
    </parent>
```

panisson:
That's all, it should be ready for 0.9.1 :fireworks:

mbastian:
Congrats! Your plugin is now available in Gephi and online at https://gephi.org/plugins/#/plugin/graphstreaming

PR #76, opened 2016-01-30
plugin node color manager migration finished

songjinghe:
this is a runnable version of this plugin, I 've test it under gephi 0.9.0 

BUT...should I checkout a new branch for this plugin? maybe I should not modify the `README` in root folder?


songjinghe:
well, this is a wrong pull request, which does not contain any of file I write, I'll send a valid new one

PR #78, opened 2016-01-31
Plugin [node color manager] migration finished

songjinghe:
well this is a runnable version tested under gephi 0.9.0. Everything seems ok except that: Should I modify the `README` file in the root directory of this repo?


mbastian:
Cool! No you shouldn't modify the README at the root of the repository. Just keep the readme in your module's folder like you already did.

modules/nodeColorManager/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+    <packaging>nbm</packaging>
+
+    <name>Node Color Manager</name>
+
+    <dependencies>
+        <!-- Insert dependencies here -->
+        <dependency>
+            <groupId>org.gephi</groupId>
+            <artifactId>tools-api</artifactId>
+            <version>0.9.0</version>
```
mbastian: You don't have to explicitly list the version here as it's inherited from the `gephi-plugin-parent`

pom.xml
```diff
@@ -12,6 +12,8 @@
     <!-- List of modules -->
     <modules>
         <!-- Add here the paths of all modules (e.g. <module>modules/MyModule</module>) -->
+       <module>modules/nodeColorManager</module>
```
mbastian: May I suggest to camelcase your plugin's folder name so it's `NodeColorManager`

songjinghe:
OK～so I should just close this pull request and fix the things you point out and create a new pull request again?

modules/nodeColorManager/pom.xml
songjinghe: well, what if the plugin updates faster than gephi ?

mbastian: That is unrelated. Your plugin finds its dependency to Gephi by inheriting the `gephi-plugin-parent` configuration, which defines explicitly a Gephi version (0.9.0 at the moment). When we update Gephi let's say to 0.9.1 we'll update this configuration and create a `0.9.1` version of the `gephi-plugin-parent`. Plugins will just need to upgrade their dependency setting for `gephi-plugin-parent`. At that time you'll have to update this line as well if you don't remove it.

songjinghe: seems I've got your point: for the same version of gephi, there can only be one rather than multiple versions of a plugin. eg. for gephi 0.9.0, there is only one version of plugin( Node Color Manager ).

Hmm, so I just fix it.

songjinghe:
well~ It seems a plugin do not need a version number for itself, so why should we input it when we run `mvn org.gephi:gephi-maven-plugin:generate` ? Maybe you should consider remove that option or ADD some notes to it(in the README in root directory : )

mbastian:
Like any Maven artifact a plugin needs a version number. Why would you like to remove it? That's the unique way to track which version of the plugin is installed.

computermacgyver:
I think the question was about where the version number is stored. It should be in the module's pom.xml file, but looking at this pull request I don't see it for this module. It probably would be a good idea to add.

modules/NodeColorManager/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.gephi</groupId>
+        <artifactId>gephi-plugin-parent</artifactId>
+        <version>0.9.0</version>
+    </parent>
+
+    <groupId>edu.buaa.act</groupId>
+    <artifactId>gephi-plugin-node-color-manager</artifactId>
```
computermacgyver: Usually the version number of the plugin is after this line. E.g.,
```
<version>1.0</version>
```

songjinghe:
@mbastian sorry I misunderstood your point then, and now I think I got it: you means dependency version number should not be declared explicitly, right? ( I make this mistake because I never use github's line comment feature and didn't know which line of code you were pointing then : P

thanks @computermacgyver for reminder me. And I've fix it

modules/NodeColorManager/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+        <!-- Insert dependencies here -->
+        <dependency>
+            <groupId>org.gephi</groupId>
+            <artifactId>tools-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.gephi</groupId>
+            <artifactId>visualization-api</artifactId>
+        </dependency>
+        <dependency>
```
mbastian: You have twice the `tools-api` dependency

modules/NodeColorManager/src/main/nbm/manifest.mf
```diff
@@ -0,0 +1,8 @@
+Manifest-Version: 1.0
+OpenIDE-Module-Name: Node Color Manager
+OpenIDE-Module-Short-Description: allow you share node colors among multi graph
```
mbastian: For the sake of presentation convention can you use a capital letter here ex: "Share node colors...".

modules/NodeColorManager/src/main/nbm/manifest.mf
```diff
@@ -0,0 +1,8 @@
...
+OpenIDE-Module-Long-Description: \
```
mbastian: There's something wrong with the formatting of this line as Java refuses to compile this manifest. You should probably use HTML tags like `Useful if you want to keep the same node to be the same color across diffrent graph/projects, by using a file to keep track of node colors among multi-graph. Note that the tool identify nodes by their labels, so your graph must have unique label for each node. If two or more node has same label, only one will be restored. When restore from file, only the nodes which match the labels in file would be colored, others will remain their original color. <br/>
    source code: <a href="https://github.com/songjinghe/gephi-plugins-1">https://github.com/songjinghe/gephi-plugins-1</a><br/>
    more details: <a href="https://github.com/songjinghe/gephi-plugins-1/wiki">https://github.com/songjinghe/gephi-plugins-1/wiki</a>`

modules/NodeColorManager/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+            <groupId>org.netbeans.api</groupId>
+            <artifactId>org-openide-dialogs</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.gephi</groupId>
+            <artifactId>graph-api</artifactId>
+        </dependency>
+        <dependency>
```
mbastian: I couldn't find code that requires this dependency, you can remote it

modules/NodeColorManager/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+            <groupId>org.gephi</groupId>
+            <artifactId>visualization</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.gephi</groupId>
```
mbastian: Same, unused dependency

modules/NodeColorManager/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+            <artifactId>layout-plugin</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.gephi</groupId>
+            <artifactId>ui-library-wrapper</artifactId>
+        </dependency>
+        <dependency>
```
mbastian: Same, unused dependency

modules/NodeColorManager/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+
+    <name>Node Color Manager</name>
+
+    <dependencies>
+        <!-- Insert dependencies here -->
+        <dependency>
+            <groupId>org.gephi</groupId>
+            <artifactId>tools-api</artifactId>
+        </dependency>
+        <dependency>
```
mbastian: Same, unused dependency

modules/NodeColorManager/src/main/nbm/manifest.mf
songjinghe: @mbastian thank you very much. I've made this manifest correct by adding white space before lines, and maven has no error, however, using HTML <a> tag result like this in gephi.
plugin panel
after you click the link, it result like this.
result
so I would prefer not using HTML tags.

songjinghe:
@mbastian well... fix again~ this is first time I've submit PR in github & cooperate with others on one project. Thank you for nice/kind suggestions! :+1:

mbastian:
You're welcome, I'm happy to help! I've tested your plugin and I have some difficulties with the file entry. If I just enter a filename (e.g. "foo") I get an exception that says "Permission Denied". You're not setting up what the file is relative too so that may be different between Windows or Mac OS X for example. In my case I'm on Mac OS X and I think the file is relative to the Application bundle, which we shouldn't write into. Then, I try to put an absolute file name such as `/Users/mbastian/foo` but I get an error that the filename is invalid and contains bad characters.

One possibility with this issue is to write your files in the Gephi configuration folder. We do that for layout presets for instance (XML files).

The following code lookup the `nodecolormanager` folder in the config folder and creates it if necessary. Then it create the `File` object relative to that folder.
```
FileObject folder = FileUtil.getConfigFile("nodecolormanager");
if (folder == null) {
   folder = FileUtil.getConfigRoot().createFolder("nodecolormanager");
}
File presetFile = new File(FileUtil.toFile(folder), fileName);
 ```

If you go that road, you'll need to add this extra dependency:
```
<dependency>
       <groupId>org.netbeans.api</groupId>
       <artifactId>org-openide-filesystems</artifactId>
</dependency>
```

songjinghe:
@mbastian well, I tested it under ubuntu linux 14.04.3, under which it would write file to the directory which start the java/Gephi process, say $PWD (because I usually start gephi from command line ), sorry I even didn't think this could be a problem under MacOS/Windows. I think it would be better to use a file choser, which will be more friendly to users.

mbastian:
Congrats and thank you @songjinghe! Your plugin is now available in Gephi.

PR #80, opened 2016-02-05
Graphviz layout plugin

chunky:
Initial pull request for 0.9-style graphviz layout plugin.

The marketplace plugin before was this: https://marketplace.gephi.org/plugin/graphviz-layout/


modules/GraphvizLayout/nb-configuration.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mbastian: I think this file is superfluous and only netbeans related. Can you remove it from the Git Index?

modules/GraphvizLayout/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>nbm-maven-plugin</artifactId>
+                <configuration>
+                    <licenseName>license</licenseName>
+                    <licenseFile>license.txt</licenseFile>
+                    <author>Gary "ChunkyKs" Briggs</author>
+                    <sourceCodeUrl>$sourcecode_url</sourceCodeUrl>
```
mbastian: Please enter this value manually, it didn't succeed to pickup the address automatically

modules/GraphvizLayout/src/main/java/org/icculus/chunky/gephigraphviz/DotProcessError.java
```diff
@@ -0,0 +1,97 @@
+/*
```
mbastian: Put license here

modules/GraphvizLayout/src/main/resources/org/icculus/chunky/gephigraphviz/Bundle.properties
```diff
@@ -0,0 +1,45 @@
+# To change this template, choose Tools | Templates
+# and open the template in the editor.
+OpenIDE-Module-Display-Category=Layout
+OpenIDE-Module-Long-Description=\
```
mbastian: Please put everything on one line and remove the "\". Otherwise it won't look good on the marketplace.

modules/GraphvizLayout/src/main/resources/org/icculus/chunky/gephigraphviz/Bundle.properties
```diff
@@ -0,0 +1,45 @@
...
+    A simple plugin that calls graphviz at the command line and uses it to lay out a graph. \
+    Graphviz has a selection of very high-quality layout algorithms available.
+OpenIDE-Module-Name=GraphvizLayout
```
mbastian: This is the "branding" name of your plugin, so I would rather recommend "Graphviz Layout"

modules/GraphvizLayout/src/main/java/org/icculus/chunky/gephigraphviz/GraphvizLayout.java
```diff
@@ -0,0 +1,314 @@
...
+        final StringBuffer dotfile = new StringBuffer();
+
+        dotfile.append("digraph g {\n");
+        dotfile.append("layout = \"").append(this.algoName).append("\";\n");
+        dotfile.append("rankdir = \"").append(this.rankDir).append("\";\n");
+        dotfile.append("overlap = \"").append(this.overlap).append("\";\n");
+        if (this.concentrate) {
+            dotfile.append("concentrate=true;\n");
+        }
+
```
mbastian: In order to guarantee consistency I would recommend to add a `graph.readLock()` here and add a `graph.readUnlock()` at line 107. Indeed, the graph could be modified between your node and edge loop, so you could have edges of new/missing nodes for instance.

modules/GraphvizLayout/src/main/java/org/icculus/chunky/gephigraphviz/GraphvizLayout.java
```diff
@@ -0,0 +1,314 @@
...
+            while ((line = outputFromGraphviz.readLine()) != null) {
+                entireOutput.append(line);
+                entireOutput.append("\n");
+//                System.out.println("line");
+            }
+            
+            // For some reason this one wasn't working
+            // Node n = graph.getNode(nodeid);
+            // ... so we map all nodes temporarily
+            final Map<Object, Node> nodeMapper = new HashMap<>();
```
mbastian: You shouldn't need that anymore, the `graph.getNode()` method should work.

modules/GraphvizLayout/src/main/java/org/icculus/chunky/gephigraphviz/GraphvizLayout.java
```diff
@@ -0,0 +1,312 @@
...
+            final String regex = "^\\s*(?<nodeid>\\S+)\\s+\\[.*?[, ]?pos=\"(?<pos>[^\"]+?)\".*?\\]";
+//            System.out.println(entireOutput);
+            final Pattern pat = Pattern.compile(regex, Pattern.MULTILINE | Pattern.CASE_INSENSITIVE | Pattern.DOTALL);
+            Matcher matcher = pat.matcher(entireOutput.toString());
+            while(matcher.find()) {
+                final String nodeid = matcher.group("nodeid");
+                
+                final Node n = graph.getNode(nodeid);
+                if(null == n) {
+                    System.err.println("Cannot find nodeid " + nodeid);
```
mbastian: It would be cleaner to use the logging system for all the `System.out` and `System.err`. You can use `Logger.getLogger("").info(msg);` or `Logger.getLogger("").warn(msg)`

mbastian:
Your plugin is now published, congratulations! Thanks and looking forward your next ideas :+1:

PR #81, opened 2016-02-15
Update for CircularLayout plugin

mgroeninger:
HI all, long time no talk!

Do I need to worry about backward compatibility of the plugin?  Does the new build pipeline replace the 0.8 version of the plugin wit one built from this source? (Since I'm getting feedback that people aren't switching yet...)  Because I haven't tested for that...

Hope things are going well,
Matt


mbastian:
> Do I need to worry about backward compatibility of the plugin? 

No you don't as the plugins repositories are separated between versions. Your plugin compatible with 0.8.2 will remain accessible.

mgroeninger:
Awesome! Thanks!

Please let me know if you have any questions or concerns.

Hope you have a great day!

modules/CircularLayout/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>nbm-maven-plugin</artifactId>
+                <configuration>
+                    <licenseName>BSDLicense</licenseName>
+                    <licenseFile>BSDLicense</licenseFile>
+                    <author>Matt Groeninger</author>
+                    <sourceCodeUrl>$sourcecode_url</sourceCodeUrl>
```
mbastian: You may want to set your GitHub fork URL here. The process that populates this at generate time didn't work apparently here

modules/CircularLayout/src/main/java/org/gephi/layout/plugin/circularlayout/circlelayout/CircleLayout.java
```diff
@@ -0,0 +1,281 @@
...
+   @Override
+   public void initAlgo()
+   {
+      this.setConverged(false);
+      this.graph = graphModel.getGraphVisible();
+      float[]        nodeCoords  = new float[2];
+      double         tmpcirc     = 0;
+      double         tmpdiameter = 0;
+      int            index       = 0;
+      int            nodecount   = this.graph.getNodeCount();
```
mbastian: You may want to wrap what follows with a graph readlock() so it guarantees consistency. For instance, nodes could be added/removed between this line and line 73, making this count wrong. You can do `graph.readLock()` and then `graph.readUnlock()` after you used the graph.

modules/CircularLayout/src/main/java/org/gephi/layout/plugin/circularlayout/circlelayout/CircleLayout.java
```diff
@@ -0,0 +1,281 @@
...
+      double         noderadius  = 0;
+      double         theta       = TWO_PI / nodecount;
+      double         lasttheta   = 0;
+      TempLayoutData posData;
+
+      Node[] nodes = this.graph.getNodes().toArray();
```
mbastian: You can safely remove the toArray() here if you use the readLock/Unlock I've mentioned earlier. This would otherwise an unnecessary copy of the nodes. Accessing nodes through `graph.getNodes()` is very fast in Gephi 0.9.0 and there's no perf. benefits of copying to an array, even if you read nodes multiple times.

modules/CircularLayout/src/main/java/org/gephi/layout/plugin/circularlayout/circlelayout/CircleLayout.java
```diff
@@ -0,0 +1,281 @@
...
+            posData.finishy = n.y();
+         }
+         posData.xdistance = (float)(1 / this.getTransitionSteps()) * (nodeCoords[0] - n.x());
+         posData.ydistance = (float)(1 / this.getTransitionSteps()) * (nodeCoords[1] - n.y());
+         n.setLayoutData(posData);
+      }
+   }
+
+   @Override
+   public void goAlgo()
```
mbastian: Same comment about locking than the `initAlgo` method

modules/CircularLayout/src/main/java/org/gephi/layout/plugin/circularlayout/circlelayout/CircleLayout.java
```diff
@@ -0,0 +1,281 @@
...
+                           "isNodePlacementTransition", "setNodePlacementTransition"));
+         properties.add(LayoutProperty.createProperty(
+                           this, Double.class,
+                           NbBundle.getMessage(getClass(), "CircleLayout.Transition.TransitionSteps.name"),
+                           TRANSITION_CATEGORY,
+                           NbBundle.getMessage(getClass(), "CircleLayout.Transition.TransitionSteps.desc"),
+                           "getTransitionSteps", "setTransitionSteps"));
+      }
+      catch (Exception e) {
+         e.printStackTrace();
```
mbastian: I would recommend to replace that with `Logger.getLogger(CircleLayout.class.getName()).log(Level.SEVERE, null, e)` so the error is propagated and will show up in Gephi's console. In general, all `System.out` and exception throwing should use this logger.

modules/CircularLayout/src/main/java/org/gephi/layout/plugin/circularlayout/dualcirclelayout/DualCircleLayout.java
```diff
@@ -0,0 +1,341 @@
...
+      setHighDegreeOutside(false);
+      setNodePlacementDirection(CircularDirection.CCW);
+      setAttribute("NodeID");
+      setNodePlacementTransition(false);
+      setTransitionSteps(100000.0);
+   }
+
+   public void setInnerNodeCount(Integer intsecondarynodecount)
+   {
+      GraphController graphController = Lookup.getDefault().lookup(GraphController.class);
```
mbastian: We need to check but calling the `graphController.getModel()` might now work when opening a .gephi project as a current workspace isn't set yet, similar to the issue described here https://github.com/gephi/gephi/issues/1373

mgroeninger:
Thanks for the notes, Mathieu! It's going to take me a few days to get back to this but I really appreciate the feedback!

mgroeninger:
I think that covered everything... Please let me know if I missed anything!

PR #84, opened 2016-02-16
Twitter Stream Plugin

totetmatt:


computermacgyver:
This is awesome, Matthieu. I think the Gephi community is going to find this plugin super helpful---I know it is something I will highlight when teaching introduction sessions.

I'm going to make a few comments/suggestions on the code to help improve it. I am not a member of the core development team; so, these are only my suggestions. However, since I have developed a few plugins for Gephi I've learned some of the common issues and hopefully this will help your plugin get launched more quickly.

Thank you again for writing this and sharing it.

modules/Twitter/README.md
```diff
@@ -0,0 +1,11 @@
+## Twitter Stream Plugin
+
+### Description
+Plugin that use the Twitter Stream Api and represent tweet as graph.
+
+2 methods of representation (network logic):
+* User Network : Do a weighter User to User network with parallel edges for RT and Mentions
+* Full Smart Netowrk : Do a full representation of User, Tweet, Hashtag, Url, Media & Symbol 
+
+This README supports Markdown, see [syntax](https://help.github.com/articles/markdown-basics/)
```
computermacgyver: This line that the readme supports markdown can be deleted.

modules/Twitter/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+        <plugins>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>nbm-maven-plugin</artifactId>
+                <configuration>
+                    <licenseName>Apache 2.0</licenseName>
+                    <author>@totetmatt</author>
+                    <authorEmail>matthieu.totet@gmail.com</authorEmail>
+                    <authorUrl>http://matthieu-totet.fr/</authorUrl>
+                    <sourceCodeUrl>$sourcecode_url</sourceCodeUrl>
```
computermacgyver: replace $sourcecode_url with the actual url of the github repository. i.e.,
``https://github.com/totetmatt/gephi-plugins/tree/twitter/modules/Twitter``

modules/Twitter/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+                    <publicPackages>
+                        <!-- Insert public packages -->
+                    </publicPackages>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+    <repositories>
+        <repository>
```
computermacgyver: I don't know why this <repositories> section is being created. It should be same to delete (or if not, we will need to figure out why it's here and get a better solution)

modules/Twitter/src/main/java/totetmatt/gephi/twitter/CredentialProperty.java
```diff
@@ -0,0 +1,71 @@
+/*
```
computermacgyver: Consider deleting this boilerplate code or changing it to something more meaningful:
```To change this license header....```

File: modules/Twitter/src/main/java/totetmatt/gephi/twitter/MainTwitterWindows.java [hunk 5]
```diff
@@ -0,0 +1,705 @@
...
+your version of this file under either the CDDL, the GPL Version 3 or
+to extend the choice of license to its licensees as provided above.
+However, if you add GPL Version 3 code and therefore, elected the GPL
+Version 3 license, then the option applies only if the new code is
+made subject to such option by the copyright holder.
+
+Contributor(s):
+
+Portions Copyrighted 2011 Gephi Consortium.
+ */
```
computermacgyver: Same here. This long introductory comment can be deleted or changed to something particular to your plugin.

modules/Twitter/src/main/java/totetmatt/gephi/twitter/MainTwitterWindows.java
```diff
@@ -0,0 +1,705 @@
...
+ * setting in the <code>@TopComponent.Registration</code> annotation:
+ * <ul>
+ * <li>rankingmode: Top left, where the appearance is.</li>
+ * <li>layoutmode: Bottom left, where the layout is.</li>
+ * <li>graphmode: Central, like the graph or data laboratory.</li>
+ * <li>filtersmode: Right, like filters or statistics.</li>
+ * </ul>
+ *
+ * @author Mathieu Bastian
+ */
```
computermacgyver: Likewise, just delete this old JavaDoc comment.

modules/Twitter/src/main/java/totetmatt/gephi/twitter/MainTwitterWindows.java
```diff
@@ -0,0 +1,705 @@
...
+        fc.setFileFilter(filter);
+        int returnVal = fc.showOpenDialog(this);
+        if (returnVal == JFileChooser.APPROVE_OPTION) {
+            File file = fc.getSelectedFile();
+            try {
+                streamer.loadTracking(file);
+            } catch (IOException ex) {
+                Exceptions.printStackTrace(ex);
+            } catch (JSONException ex) {
+                Exceptions.printStackTrace(ex);
```
computermacgyver: Rather than just printing the stack trace to the console, it would be better to use a logger. The same for the exception just above this line.

The following code will propagate the error (i.e., the user will see it in a dialogue box as well as it getting logged to the Gephi logs)
```
Logger.getLogger(MainTwitterWindows.class.getName()).log(Level.SEVERE, null, ex);
```

modules/Twitter/src/main/java/totetmatt/gephi/twitter/MainTwitterWindows.java
```diff
@@ -0,0 +1,705 @@
...
+            }
+            refreshTracking();
+        } else {
+            //log.append("Open command cancelled by user." + newline);
```
computermacgyver: You could either just remove this else statement (since it is empty) or consider replacing the commented out line with another logging statement:
```
Logger.getLogger("").info("Open command cancelled by user.");
````

File: modules/Twitter/src/main/java/totetmatt/gephi/twitter/MainTwitterWindows.java [hunk 9]
```diff
@@ -0,0 +1,705 @@
...
+        fc.setFileFilter(filter);
+        int returnVal = fc.showSaveDialog(this);
+        if (returnVal == JFileChooser.APPROVE_OPTION) {
+            File file = fc.getSelectedFile();
+            try {
+                streamer.saveTracking(file);
+            } catch (JSONException ex) {
+                Exceptions.printStackTrace(ex);
+            } catch (IOException ex) {
+                Exceptions.printStackTrace(ex);
```
computermacgyver: As above, consider using Logger here in place of both printStackTrace commands.

modules/Twitter/src/main/java/totetmatt/gephi/twitter/TwitterStreamer.java
```diff
@@ -0,0 +1,151 @@
...
+    TwitterStream twitter;
+    CredentialProperty credentialProperty = new CredentialProperty();
+    List<String> wordTracking = new ArrayList<String>();
+    Map<String, Long> userTracking = new HashMap<String, Long>();
+
+    public void addUser(String screenName) {
+        try {
+            long id = UserIdResolver.resolve(screenName);
+            userTracking.put(screenName, id);
+            System.out.println(id);
```
computermacgyver: Delete this ``System.out.println`` or replace with a Logger statement

modules/Twitter/src/main/java/totetmatt/gephi/twitter/TwitterStreamer.java
```diff
@@ -0,0 +1,151 @@
...
+        } catch (IOException ex) {
+            Exceptions.printStackTrace(ex);
```
computermacgyver: Replace
`` Exceptions.printStackTrace(ex)``
with
``Logger.getLogger(TwitterStreamer.class.getName()).log(Level.SEVERE, null, ex);``

modules/Twitter/src/main/java/totetmatt/gephi/twitter/networklogic/FullSmartNetwork.java
```diff
@@ -0,0 +1,186 @@
...
+
+    }
+
+    @Override
+    public void onStatus(Status status) {
+        processStatus(status, null);
+    }
+
+    public void processStatus(Status status, Node retweetUser) {
+        Node tweet = graphController.getGraphModel().getGraph().getNode("" + status.getId());
```
computermacgyver: I don't know how much of a problem this is, but it is possible that the graph could be modified in between some of these statements. For example the node representing the tweet could exist, but then be deleted before the createLink(user,tweet) call. To avoid this, you can add
``graphController.getGraphModel().getGraph().writeLock()``
at the beginning of the function and
```graphController.getGraphModel().getGraph().writeUnlock()``
at the end. This will prevent changes to the graph while you are updating it.

File: modules/Twitter/src/main/java/totetmatt/gephi/twitter/networklogic/UserNetwork.java [hunk 13]
```diff
@@ -0,0 +1,103 @@
...
+        String originScreenName = status.getUser().getScreenName().toLowerCase();
+
+        // For each mention
+        for (UserMentionEntity mention : status.getUserMentionEntities()) {
+
+            String targetScreenName = mention.getScreenName().toLowerCase();
+            // Avoid self-loop
+            if (!originScreenName.equals(targetScreenName)) {
+
+                // Create or get the nodes
```
computermacgyver: The idea of locking the graph also applies within this if statement

modules/Twitter/src/main/nbm/manifest.mf
```diff
@@ -0,0 +1,5 @@
+Manifest-Version: 1.0
+OpenIDE-Module-Name: Twitter
+OpenIDE-Module-Short-Description: Twitter Stream
+OpenIDE-Module-Long-Description: Get Twitter Stream displayed
```
computermacgyver: These are the values that will appear for your plugin with Gephi and also within the plugin marketplace. I guess they are all right, but you might consider adding a little more detail (particularly to the Long Description). You might also make the Module Name a little more specific. For example,
```
OpenIDE-Module-Name: TwitterImporter
OpenIDE-Module-Short-Description: Plugin to create a graph from the Twitter Streaming API
OpenIDE-Module-Long-Description: <p>This plugin connects to the Twitter Streaming API and represents tweets as graph.</p><p>It includes two possible ways to represent a graph:<ul><li>User Network: A weighted User to User network with parallel edges for RT and Mentions</li><li>Full Smart Network: A full representation with nodes for Users, Tweets, Hashtags, Urls, Media and Symbols</li></p>

File: pom.xml [hunk 15]
```diff
@@ -12,6 +12,7 @@
     <!-- List of modules -->
     <modules>
         <!-- Add here the paths of all modules (e.g. <module>modules/MyModule</module>) -->
+        <module>modules/Twitter</module>
```
computermacgyver: I don't know of any other plugin called "Twitter" so this could work, but otherwise consider being a bit more specific. For example, "TwitterStreamingImporter" (and updating the folders, etc. to use this new name)

modules/Twitter/src/main/java/totetmatt/gephi/twitter/utils/UserIdResolver.java
```diff
@@ -0,0 +1,27 @@
...
+        while (els.hasNext()) {
+            Element el = els.next();
+            if (!el.attr("data-user-id").isEmpty()) {
+                return Long.parseLong(el.attr("data-user-id"));
+            }
+        }
+
+        return 0;
+    }
+}
```
computermacgyver: Out of curiosity, why are you directly accessing the HTML of a Twitter page rather than using the REST API? The [GET users/lookup](https://dev.twitter.com/rest/reference/get/users/lookup) endpoint is designed to for this username to numeric userid conversion and can also handle batches of up to 100 usernames (rather than requiring a separate request for each username)

totetmatt: Long time that I didn't look actually at the API,and at that time it wasn't so permissive (drastic call limit mainly). One for the reason I've done this way was because you don't need to "connect" and it doesn't consume API limit. I'll have a look on the GET users/lookup and the interface with Twitter4J to check if it would be effectively better to use directly the API.

modules/TwitterStreamingImporter/src/main/java/totetmatt/gephi/twitter/MainTwitterWindows.java
```diff
@@ -0,0 +1,664 @@
...
+ * Window. Then, one can customize the default position by changing the mode
+ * setting in the <code>@TopComponent.Registration</code> annotation:
+ * <ul>
+ * <li>rankingmode: Top left, where the appearance is.</li>
+ * <li>layoutmode: Bottom left, where the layout is.</li>
+ * <li>graphmode: Central, like the graph or data laboratory.</li>
+ * <li>filtersmode: Right, like filters or statistics.</li>
+ * </ul>
+ *
+ * @author Mathieu Bastian
```
mbastian: I certainly can't get any credits for this class :)

modules/TwitterStreamingImporter/src/main/java/totetmatt/gephi/twitter/networklogic/Networklogic.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+    public void setTrack(String[] track){
+        this.track = track;
+    }
+    
+    // This is call for each tweet received, it *needs* to be defined afterward.
+    public final void onStatus(Status status) {
+        graphController.getGraphModel().getGraph().writeLock();
+        processStatus(status);
+        graphController.getGraphModel().getGraph().writeUnlock();
```
mbastian: It could be good to wrap this into a try/finally so if something fails with `processStatus` the graph is unlocked anyway.

mbastian:
One general comment about your calls to `graphController.getGraphModel().getGraph()`, I would recommend to proceed differently: obtain the current workspace at the top of your funnel in the UI and then pass down a reference to `GraphModel`. That will avoid a lot of calls to `getGraphModel()` which isn't inefficient but it's still a lot of redundant work with locking/synchronisation consequences. I haven't tested your plugin yet but I imagine that it's possible for the user directly or another plugin to switch workspaces while your plugin is operating and that's dangerous as suddenly the `getGraphModel` returns a different graphModel.

Also, it would be awesome if you could attach one or two screenshots to your plugin. It's really simple, just put images into the `src/img` folder and the gephi-maven-plugin will take care of the rest so they are displayed on the plugins website.

modules/TwitterStreamingImporter/src/main/resources/totetmatt/gephi/twitter/Bundle.properties
```diff
@@ -0,0 +1,46 @@
+# To change this license header, choose License Headers in Project Properties.
+# To change this template file, choose Tools | Templates
+# and open the template in the editor.
+
+CTL_SimpleAction=Twitter Stream
+CTL_SimpleTopComponent=Twitter Stream
+HINT_SimpleTopComponent=WHEEEEEEE
```
mbastian: That's what appears in the tooltip in the component and the copy can be improved I think :)

modules/TwitterStreamingImporter/src/main/resources/totetmatt/gephi/twitter/Bundle.properties
```diff
@@ -0,0 +1,46 @@
...
+OpenIDE-Module-Display-Category=Plugin
```
mbastian: You already put this information in the `manifest.mf` file and you can remove this, as well as the lines below.

mbastian:
Testing your plugin now. Unfortunately on Mac OS X there are issues when windows are overlapping with the Graph window and you obtain something like this:
screen shot 2016-02-19 at 10 24 07 pm
I don't really have a workaround to suggest :/ and would recommend using another mode than the graphmode at the moment. I know that forces you to probably reshuffle your UI and fit into a smaller real estate in either `layoutmode`, `rankingmode` or `filtersmode`.

mbastian:
There seems to be a little layout issue with some hidden element on Mac OS:
screen shot 2016-02-19 at 11 14 00 pm

mbastian:
Cool, not an issue but a suggestion regarding edge types. I think there's a nice opportunity to leverage the multi-graph capabilities. Right now all edges are created equal but you could have typed edges (mention, follow etc..). You can do so by providing a type integer to the `newEdge()` method. This type can remain an integer if you like (e.g. 1 = follow etc) but the best is to associate this with a label for relationship, which is something gephi can do. If you call `int type = graphModel.addEdgeType("Follow")` you can use this integer without having to worry about it's value. If you call `addEdgeType()` repeatatly it will just return the same so you could streamline this into the `newEdge()` function if you like, e.g. `newEdge(source, target, graphModel.addEdgeType("Follow"), 1.0, true)`. although it's not as fast as caching the types.

eduramiba:
Hi Totet,
Are the changes complete?

Thanks,
Eduardo

totetmatt:
hi,
if eveything is right on gephi team side, no more change to complete.
But as explain on gephidev chat, I will be unable to support the plugin for the moment as im "away from internet" until mid april. Therefore I would like to wait if possible, before merging so that I can fully support feedback from user when I will be back.

eduramiba:
Hi Totet, pretty awesome work!

Can I suggest a few small improvements before merge?

One would be adding a 'Type' string column or similar to nodes, so data can be filtered more easily.
The other is a better initial random position, similar to the positions created by data laboratory, also a initial size of 10.

I created a pull request in your branch with these changes + a few details https://github.com/totetmatt/gephi-plugins/pull/1

eduramiba:
Plugin available, congrats!

Issue #86, opened 2016-02-17
MultiMode Projection deletes all edges


judithmp:
The MultiModeNetworksTransformation Plugin in 0.9/0.9.1 doesn't create new edges for the one-mode projection - instead all edges of the network are gone and the graph becomes directed after projection (happening at step 4.3. of this tutorial: http://www.martingrandjean.ch/gephi-introduction/).
![gephi_multimode_before](https://cloud.githubusercontent.com/assets/5262148/13127126/19ec146e-d5ce-11e5-820d-f81f8f277356.png)
![gephi_multimode_after](https://cloud.githubusercontent.com/assets/5262148/13127133/24850f66-d5ce-11e5-8902-194a62bbbb34.png)

totetmatt:
**Error Thrown : **
```
java.lang.IllegalArgumentException: Can't add a directed edge to an undirected graph
	at org.gephi.graph.impl.UndirectedDecorator.addEdge(UndirectedDecorator.java:43)
	at cz.cvut.fit.gephi.multimode.LongTaskTransformation.run(LongTaskTransformation.java:178)
[catch] at org.gephi.utils.longtask.api.LongTaskExecutor$RunningLongTask.run(LongTaskExecutor.java:274)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

**Workaround before bug fix **: Save as Gexf, re-open but as "Graph Type : Mixed".

eduramiba:
Closing as this was fixed

PR #87, opened 2016-02-22
Multimode network

totetmatt:
Bug fix for https://github.com/gephi/gephi-plugins/issues/86


mbastian:
Plugin updated, thanks for the patch!

PR #89, opened 2016-02-28
Loxa Web site export

jorgaf:
Hello,

I finish to adapt my Gephi plugin, please add to the repository.

Greetings.


eduramiba:
Thanks Jorge,
Your plugin is now published.

I made a few changes in the pom.xml, removed some code comments and added a plugin image. Please update your code with these changes for later pull requests, as it seems things like the metrics panel will be updated in the future.

PR #99, opened 2016-04-04
Isometric layout

alexandrebarao:
Pull request from my branch isometric-layout.
Best regards to the team and special thanks to Eduardo.
Alexandre Barão


modules/IsometricLayout/src/main/java/org/alexandrebarao/isometriclayout/IsometricLayout.java
```diff
@@ -0,0 +1,442 @@
...
+        
+        
+        graph.readUnlock();
+        endAlgo();
+        
+        
+    }
+
+    
+    public double getDataLaboratoryValue(Node n, Column col) {
```
eduramiba: Here you could check if the column is a subclass of ```Number``` Then use doubleValue() on the value.
Also, parsing Strings to Double might throw a parse exception. I suggest catching and ignoring that exception since that's not very likely to happen.

alexandrebarao: Done. Pulling next.

eduramiba:
Congrats! Your plugin is now available :)

I changed one System.out to use java Logger. Please include the changes into your branch for future releases.

Sorry for the delay!

PR #100, opened 2016-04-04
Network splitter 3D

alexandrebarao:
Pull request from my branch network-splitter-3d.
Best regards to the team and special thanks to Eduardo.
Alexandre Barão


modules/NetworkSplitter3D/src/main/java/org/alexandrebarao/networksplitter3d/NetworkSplitter3D.java
```diff
@@ -0,0 +1,502 @@
...
+                         
+            
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+
+        return properties.toArray(new LayoutProperty[0]);
+    }
+
+     public double getDataLaboratoryValue(Node n, Column col) {
```
eduramiba: Here you could check if the column is a subclass of ```Number``` Then use doubleValue() on the value.
Also, parsing Strings to Double might throw a parse exception. I suggest catching and ignoring that exception since that's not very likely to happen.

modules/NetworkSplitter3D/src/main/java/org/alexandrebarao/networksplitter3d/NetworkSplitter3D.java
```diff
@@ -0,0 +1,502 @@
...
+import org.gephi.layout.spi.LayoutProperty;
+import org.openide.util.Lookup;
+
+
+public class NetworkSplitter3D implements Layout {
+
+    
+    // Network Splitter 3D Data Structures & Settings
+    
+    ArrayList nodes3D;
```
eduramiba: I suggest using generics and privates wherever possible.

modules/NetworkSplitter3D/src/main/java/org/alexandrebarao/networksplitter3d/NetworkSplitter3D.java
```diff
@@ -0,0 +1,502 @@
...
+                   if ( col.getTypeClass()== Double.class ) { 
+                        Double colString = (Double) (n.getAttribute(col));
+                        z = colString.doubleValue();
+                   }
+                   
+                   return z;
+    }
+    
+     
+      public Node searchNode(Graph graph, int id) {
```
eduramiba: You probably can replace all this code with https://gephi.org/gephi/0.9.1/apidocs/org/gephi/graph/api/Graph.html#getNode-java.lang.Object-

modules/NetworkSplitter3D/src/main/java/org/alexandrebarao/networksplitter3d/NetworkSplitter3D.java
```diff
@@ -0,0 +1,502 @@
...
+                          }
+                      }
+
+                    ++nodeK;
+                }
+        }
+        return n;
+    }
+    
+      public Node3D searchNode3D(int id) {
```
eduramiba: This could probably be simplified by using a ```HashMap<Integer, Node3D>``` instead of a list.

modules/NetworkSplitter3D/src/main/java/org/alexandrebarao/networksplitter3d/NetworkSplitter3D.java
alexandrebarao: done.

modules/NetworkSplitter3D/src/main/java/org/alexandrebarao/networksplitter3d/NetworkSplitter3D.java
alexandrebarao: done

modules/NetworkSplitter3D/src/main/java/org/alexandrebarao/networksplitter3d/NetworkSplitter3D.java
alexandrebarao: Justified in commit message.

modules/NetworkSplitter3D/src/main/java/org/alexandrebarao/networksplitter3d/NetworkSplitter3D.java
alexandrebarao: Agree. However i am using now generic data structures as suggested and i have optimized some code loops. Using a hashmap will significativelly change the tested code. I kindly request permission to keep as it is because the solution is stable and the code is clean.

eduramiba:
Congrats! Your plugin is now available :)

I changed one System.out to use java Logger. Please include the changes into your branch for future releases.

Sorry for the delay!

PR #113, opened 2016-05-29
Updated event graph layout for gephi 0.9.

WouterSpekkink:
I noticed that floats and Float objects can no longer be cast to Double objects, and that doubles or Double objects cannot be cast into Float objects. This was possible with the previous API of Gephi, as far as I know. Now Gephi will throw an error of such casting is attempted. The same happens with the GeoLayout plugin I have noticed.


eduramiba:
Hi,
The cast is no longer possible because Gephi used to store primitive numbers (float, double) but now stores Float or Double for example, so autoboxing is not possible.

eduramiba:
Sorry for the delay, I created a pull request on your branch so we can finally merge this :)

WouterSpekkink:
Hi Eduardo,

Thanks for picking up the work on the plugins. I'm a bit new to these merger issues. Is it just a matter of me going through your edits and merge your pull requests on my branch, or is there more I should do?

Also, I think I fixed the casting issue before right? I remember that after reporting the issue for GeoLayout, I came up with another solution that doesn't use casting with primitives.

Best,

Wouter

eduramiba:
Hi Router,
Yeah, the fix for casting numbers is perfect. Also you only need to merge my pull requests and your plugin will be available shortly when I merge this.

eduramiba:
Wouter* sorry bad auto correct haha

WouterSpekkink:
Router, that's a new one :) No worries, you should hear the various
pronunciations of my name that I've heard. Thanks Eduardo!

Wotuer

2016-09-11 14:04 GMT+01:00 Eduardo Ramos <notifications@github.com>:

eduramiba:

PR #118, opened 2016-05-29
Mds metric

WouterSpekkink:
Updated this plugin for Gephi 0.9. It makes sense to review the MDS Layout plugin first, because that plugin is needed to use the results of the MDS Statistics plugin to create layouts.


WouterSpekkink:
Having an issue with this plugin add the moment. Probably to do with the mdsj.jar.

PR #120, opened 2016-06-01
A filter from file plugin

ypanagis:
I created a plugin for users to be able to keep only those nodes that have labels from a list of names, provided in a text file. The filter is available under **Filters>Attributes** as a new entry outside all categories called _Filter by list_. 

The user can then create a ``.txt`` file, with one label per row, locate the file in the filesystem from the filter panel by pressing **Choose file** and then press **Filter** to see the results

An example such file would contain the entries:

```
Gavroche
Valjean
```
for the Miserables' network and would then keep only the nodes of the two heroes OR those nodes plus their immediate neighbors if the user checks the **Include neighbors** option in the filter


eduramiba:

PR #121, opened 2016-06-02
MultiGravity ForceAtlas2

suryanash:
I made a modification of ForceAtlas2 using multiple sources of gravity. Graphs are expected to have an attribute which is "gravity_x" and "gravity_y" which need to be doubles, that give approximate locations of sources of gravity. The distance between the sources of gravity can be modified using the parameters for the layout. The main benefit of this algorithm is that graphs can be coerced into certain shapes based on some data related to the data set.


eduramiba:
Sounds cool! Could you please add some sample images to your plugin? See https://github.com/gephi/gephi-plugins/tree/geolayout-plugin/modules/GeoLayout/src/img for example

suryanash:
Have now added some images of the modification in use. I have also added images of what the layouts looked like using just FA2, to compare to how they look with MGFA2.

suryanash:
I have removed the previous images and added ones that I believe demonstrate the use of this layout better.

eduramiba:
Hi,
I created a pull request in your branch that fixes some parts of the code. Please accept it so we can merge this.

eduramiba:
Merged in commit d4a43cbb4a390f35ac923f68b7666b624b041eac

eduramiba:

PR #131, opened 2016-09-18
Update multimode network plugin

joancf:
I did some updates to the multimode network plugin.
What I did was to add the cosine similarity and a threshold. 
The threshold is used to avoid the creation of low weight edges, making the resulting graph more light weight
The cosine similarity (between two nodes) is scaled between -100.0 and 100.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,6 +22,7 @@
         <module>modules/GraphvizLayout</module>
         <module>modules/GraphStreaming</module>
         <module>modules/DesktopStreaming</module>
+        <module>modules/StreamingAPI</module>
         <module>modules/StreamingImpl</module>
         <module>modules/JettyWrapper</module>
         <module>modules/StreamingServer</module>
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,6 @@
         <module>modules/GraphvizLayout</module>
         <module>modules/GraphStreaming</module>
         <module>modules/DesktopStreaming</module>
-        <module>modules/StreamingAPI</module>
         <module>modules/StreamingImpl</module>
         <module>modules/JettyWrapper</module>
         <module>modules/StreamingServer</module>
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .DS_Store
 **/target/**
 /target/
+/nbproject/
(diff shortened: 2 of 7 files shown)


pom.xml
```diff
@@ -22,7 +22,6 @@
         <module>modules/GraphvizLayout</module>
         <module>modules/GraphStreaming</module>
         <module>modules/DesktopStreaming</module>
-        <module>modules/StreamingAPI</module>
```
eduramiba: This must be an error. Also, it would be preferrable to make a pull request from a branch that only contains your plugin, instead of master-forge.

eduramiba:
Hi, you will need to increase the plugin version in manifest.mf file in order for autoupdate to work.

pom.xml
joancf: Eduardo, 
As I modified the plugin, I did not know how to get the simple plugin, maybe I had to do git on the folder, (it was difficult also to find which was the branch  because it's not on master.) I'ts true that I could not compile the full project becasue the StreamingApi was giving me errors. and I removed it from Netbeans... I'll try to do a pull request of the single module
By the way I could not run Gephi from maven, I tried the module on a regular install of gephi.

eduramiba: Oh, I didn't notice you are not the original developer of the plugin. 
In that case, you should work with the original developer branch: https://github.com/totetmatt/gephi-plugins/tree/multimode-network and then create a pull request there with your changes. Then the original developer should merge your changes and create another pull request here.

joancf:
But he is not the original author of the plugin... that should be updated when installing the plugin in 9.0.1. 
It was not easy to find the code... I'll try to talk with him soon.

eduramiba:
Oh sorry that's right, the author is different at https://marketplace.gephi.org/plugin/multimode-networks-transformations-2 ...

Then I can review your changes to the plugin, but you should still use the branch at https://github.com/totetmatt/gephi-plugins/tree/multimode-network instead of master-forge because that's what was merged into master-forge with commit https://github.com/gephi/gephi-plugins/commit/473b67c5e9fff10e66f4f98a5282d257caa8317e (also, this way you will be able to run the plugin without problems). Also please try to talk with @totetmatt about the changes.

totetmatt:
Hey, sorry for delay. Yes i'm not the original creator of the script, I just performed the transition from 0.8 to 0.9 at that time because I needed the plugin. 

So I don't know what's needed on my side, but if it's working with more improvment, you can actually rebranch to joancf branch.

eduramiba:
Great, merging this then.

PR #141, opened 2017-01-29
Add HTTP Graph for Gephi 0.9

phreakocious:
Now works with 0.9.  Also new support added for receiving data from a separate Chrome extension.


eduramiba:
Thanks for the request! I will check this soon.

phreakocious:
Could this get looked at soon, please?

eduramiba:
Maybe I can check this later today

eduramiba:
Looks good to me! Only problem is you should remove `<version>0.9.1</version>` from your module pom.xml dependencies to `lib.validation` and `io-importer-api`. This is not necessary because dependency versions are managed by parent pom.

Please do this so I can marge.

phreakocious:
Thanks, @eduramiba .. Done.

eduramiba:
Your plugin is now available in the update center :smiley:

PR #146, opened 2017-02-24
Map of countries branch

eduramiba:
Fix https://github.com/gephi/gephi-plugins/pull/145


eduramiba:
@seinecle Merged in master-forge, will publish updates soon

eduramiba:
@seinecle Available now in update center (probably considered a different plugin due to the artifactid change). I recommend to remove the old one first if installed.

eduramiba:
@seinecle Mmm I notices it actually causes an exception adding the column... we will need to fix this 😮 

The problem is it checks for the wrong table: https://github.com/gephi/gephi-plugins/pull/146/commits/58d5307e64632e89e300a51e878e9d0e78a458a8#diff-555649f82d8517bcb2675bbd8c3806ecR110

seinecle:
I'll revise it on Monday

modules/MapOfCountries/src/main/java/net/clementlevallois/mapofcountries/GeoLayout.java
```diff
@@ -107,6 +107,10 @@ public void goAlgo() {
                 graphModel.getNodeTable().addColumn("background_map", Boolean.class);
             }
             
+            if (!graphModel.getEdgeTable().hasColumn("background_map")){
+                graphModel.getNodeTable().addColumn("background_map", Boolean.class);
```
seinecle: Line 111 is a bad copy paste. Shd be:
graphModel.getEdgeTable()

modules/MapOfCountries/src/main/java/net/clementlevallois/mapofcountries/GeoLayout.java
```diff
@@ -156,7 +160,7 @@ public void goAlgo() {
             for (Edge e : graph.getEdges().toArray()) {
                 Node nodeSource = e.getSource();
```
seinecle: Lines 157-158 can be deleted.

PR #154, opened 2017-03-08
changed some class names to avoid collision with other plugins.

seinecle:
tested when installed with the Excel/Csv importer plugin, runs fine now.

removed a System.out.println, too.


eduramiba:
Hi,
Can you push a new commit to increase the plugin version?

seinecle:
damn forgot that. Doing it now.

eduramiba:
The update is now available

seinecle:
Fantastic. Thanks a lot.

PR #159, opened 2017-05-12
Allow reconnecting clients without needing to enter settings again

### LINKED ISSUE #158: Cannot restart stream client
You can start/stop the server, but cannot do so with the client. Would be useful to be able to do this.

virtuald:
- Fixes #158


eduramiba:
Merged thanks! I will release this new version soon :)

PR #163, opened 2017-06-19
new layout plugin

weihs:
make a modification on expand algorithm to enable users to expand or contract part of the graph.


eduramiba:
Thanks for the request, will review as soon as I can

eduramiba:
Merged! Your plugin will be available soon

eduramiba:
I fixed $sourcecode_url in the pom, but please change it in your branch too

eduramiba:
Your plugin is now available :)

PR #164, opened 2017-07-08
scale layout plugin update

weihs:
Fix the sourcecode URL in pom.xml.
Add my institute name after my name in pom.xml, According to college policy, I should mention my college when I release any part of my thesis.
Add batch operation, now users can manipulate multiple parts of the graph simultaneously.


eduramiba:
Hi,
Can you restore nbactions.xml (used for running easily from netbeans).

eduramiba:
In batch mode, each part is scaled locally using only its own part mean, is that the desired behavior?

Maybe the default values could be start=1, end=1 so it's less confusing

weihs:
Yes, that is my desired behavior. I will modify the default values and description.
I will also restore the nbactions.xml, although I am confused about this file. It would be appreciated if you could tell me or offer me a link about how to restore this.
Considering I am on holiday now, maybe I cannot fix them until I come back. But anyway, I will fix them.
Thanks for your information.

eduramiba:
Hi,
nbactions is not necessary for compiling, but is a file that offers an easy way of running the plugin (right click + run) without having to write a long command. You can just restore the version from master.

weihs:
Hi,
I have already added the nbaction.xml and modified default values of parameters to avoid confusion.

eduramiba:
Hi, thanks. Actually, the pom version was fine, you should increase it in order for updates of the plugin to be available. You can set anything > 1.0.0

weihs:
Hi，
Thanks for your reminding, I have already increased the version code. It is the first time for me to update a plugin, sorry for my carelessness.

eduramiba:
Merged, thanks!
The update will be available shortly

PR #171, opened 2017-09-24
Update version to Gephi 0.9.2

eduramiba:


eduramiba:
This should publish all plugins rebuilt for 0.9.2

eduramiba:
We will actually need to update version references on each module

Issue #168, opened 2017-09-12
[MultimodeNetworksTransformationPlugin] All edges get lost when transforming 2-mode to 1-mode


michihenninger:
The new version of the MultimodeNetworksTransformationPlugin 1.0.1 (released at Aug 10) seemes to have introduced a bug. We are not able to transform a 2-mode network into a one-mode network. After the transformation all edges are disapprared. (But removing edges was not selected)

I have a workaround (install older version manually) but I would appreciate using the plugin manager and having a working version.

joancf:
Yes I found that error, it happens in some networks, not always.
 I did some changes to that plugin some time ago, but I can't get the code of the current version of the plugin for gephi 9.2

eduramiba:
Related: https://github.com/gephi/gephi/issues/1854

mbastian:
Looks like this was merged already. Thanks!

PR #173, opened 2017-11-24
Multimode

joancf:
#168


eduramiba:
Hi,
Thanks for the pull request! I will take a look as soon as possible. But please you should only change the version and author in the pom.xml (and add needed dependencies). No other changes should be necessary (for example the plugin config, and you don't need to repeat `gephi.version`, as that dependency version is already managed in parent pom)

eduramiba:
Related issue: https://github.com/gephi/gephi/issues/1854

eduramiba:
Thanks, I finally merged this but had to add several fixes and sorted out the code a bit.

See https://github.com/eduramiba/gephi-plugins/commit/083b56b04fd1bb40024f0badf206eb981a11af56

The problem with not creating any edges (https://github.com/gephi/gephi/issues/1854) was for directed graphs due to the changes in https://github.com/gephi/gephi-plugins/pull/131 where the condition of neighbors for second matrix was changed from being neighbor to getting the edge weight, but in a directed graph it won't work because `getEdge` takes into account edge direction.

Now it will treat any graph as undirected unless directed option is checked.
It will also sum the weights in case of parallel edges.

PR #176, opened 2018-01-20
Twitter

totetmatt:
New Features : 
 - Add users from twitter list (https://github.com/totetmatt/gephi-plugins/issues/8)
 - Add Track by Location


eduramiba:
Thanks for the request! I will review as soon as possible

eduramiba:
Merged, thanks! Your plugin should be available soon.
Exceptions can happen if lat long floats are not written correctly, it would be great to catch these cases and show a friendly message though.

eduramiba:
The new version is now available!

PR #180, opened 2018-03-03
Twitter Stream Importer 1.3.0 : Emoji Network

totetmatt:


eduramiba:
Thanks for the request! I expect to merge it this weekend.

totetmatt:
Don't worry, no need to rush for this ;-)

eduramiba:
Merged, awesome! I just changed the Manifest-Version to 1.0 as it should never change, it's the version of the manifest file type.

eduramiba:
The new version should be available soon

PR #183, opened 2018-05-18
Newman Girvan Plugin

jSh4rk:
This plugin uses the algorithm of Newman-Girvan to cluster a given graph. The plugin implements the shortest path betweeness and modularity to find the best clustering. It extends the idea of Newman-Girvan and can respect edge types and multi edges between tweo nodes. 

It was created as a part of a practical internship at the university of hagen at the department of theoretical computer science.


eduramiba:
Thanks for the request! I will review it as soon as possible.

eduramiba:
Merged, thanks! Your plugin should be available soon

PR #184, opened 2018-06-02
Image preview plugin to 0.9.X

totetmatt:
== Transition to 0.9.X ==
Main work was to make the plugin transition to 0.9.X. 

Main issue was the `renderImageProcessing` (`ImageNode.java`) function as now the default rendering engine is `G2DTarget`. It makes the old Processing trick unusable anymore. Therefore, it's now using fully the `awt.Graphic2D` to render the image on preview and PNG rendering.

No big issue with PDF and SVG rendering.

== UX Changes ==
* Adding a `Scaling factor` that keep the size aspect but can resize the images only.
* The Default image directory used is your home/Document, if it's changed, the plugin keep the last Directory used even after a restart
* Alpha doesn't have any effect (didn't find correct method with Graphic2D)
* Some error handeling

== To Keep in Mind ==
* Didn't tried on lare graph (Maybe some Reuse of similar images could help ?)

Open to enhancement proposition before merging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .DS_Store
 **/target/**
+/target/
--- a/modules/gephi-plugin-image-preview/pom.xml
+++ b/modules/gephi-plugin-image-preview/pom.xml
@@ -27,6 +27,7 @@
     <dependency>
       <groupId>org.netbeans.api</groupId>
       <artifactId>org-openide-util-lookup</artifactId>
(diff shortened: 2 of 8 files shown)


eduramiba:
Thanks for the pull request! I hope to have some time soon to check all your requests.

eduramiba:
Merged with improvements, thanks!

PR #186, opened 2018-06-20
Katz Centrality

MonikaNowakk:
Katz centrality of a node is a measure of centrality in a network.
Link:
https://en.wikipedia.org/wiki/Katz_centrality


eduramiba:
Hi, thanks for the pull request, I will review it as soon as possible.
Could you please remove .idea folder and add it to .gitignore?

eduramiba:
Also, your module is missing pom.xml. Please correct it so its format is the same as other plugins.

mbastian:
Closing as we have merged another katz centrality plugin, see #231. It probably doesn't make sense to have two but we appreciate the effort!

PR #207, opened 2019-05-13
Twitter 1.4.0

totetmatt:
Release 1.4.0

# Start Message
A message that invite to start a new project in case Gephi is in a state where no project is currently on going. This message disappear when this condition is met.

# Multiple Input
For word and user, it's now possible to enter multiple entries in one by using the comma ',' to separate entities:

# Quotes, Mentions & Retweet
Twitter API & Twitter4J has changed for quite a while now their way to represent "user interaction". As now the differentiation are stable, the UserNetwork and the FullNetwork has been updated to reflect better this representation from twitter.

### DIFF
--- a/modules/TwitterStreamingImporter/src/main/nbm/manifest.mf
+++ b/modules/TwitterStreamingImporter/src/main/nbm/manifest.mf
@@ -1,4 +1,4 @@
-Manifest-Version: 1.3
+Manifest-Version: 1.4
 OpenIDE-Module-Name: TwitterStreamingImporter
 OpenIDE-Module-Short-Description: Realtime Twitter Graph Importer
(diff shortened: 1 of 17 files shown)


modules/TwitterStreamingImporter/src/main/nbm/manifest.mf
```diff
@@ -1,4 +1,4 @@
-Manifest-Version: 1.3
+Manifest-Version: 1.4
```
eduramiba: I don't think the manifest version should be ever different than `1.0` though. It's not the project version but the file format version.

eduramiba:
Will merge this soon as it's just an update, I guess it works fine :)

modules/TwitterStreamingImporter/src/main/nbm/manifest.mf
totetmatt: Oh stupid I am :D I'll change this

PR #210, opened 2019-06-03
Twitter

totetmatt:
Little bug fix that avoid to crash the streaming if it's stopped and  launched a second time with the new projection


eduramiba:
Nice!

PR #211, opened 2019-06-24
Twitter browser importer - Plug in

mjorda98:
Plugin that given a tweet link it opens a browser, based on chromium, and creates a graph of the comments, the likes or the retweets. All the information to create this graph is taken from the source html code of twitter. 
Prerequisites:
-	Google chrome 73.0
-	A twitter account (to create the likes and the retweet graphs)

We are a group of four students of the university UPF (Universitat Pompeu Fabra), that our final project for a subject is this plug-in. If we need to improve or change something in the code to get the plug-in approved, please tell us and we will change it without any problem.


mjorda98:
Also inside the bin folder of gaphi it has to be a chromium file.
You can download it from: 
https://chromedriver.storage.googleapis.com/index.html?path=73.0.3683.68/

eduramiba:
Thanks for the request, we will test it and merge.

PR #220, opened 2020-05-31
Twitter Streaming Importer 1.4.2

totetmatt:
TSI_142

* Add possibility to filter per language (Twitter API Feature) : https://developer.twitter.com/en/docs/tweets/filter-realtime/guides/basic-stream-parameters

* Add "Random Sample" that will use the sample Stream API (Twitter API Feature) :  https://developer.twitter.com/en/docs/labs/sampled-stream/overview

* Review  Save & Load filter configuration in json


eduramiba:
Great! Thank you Matthieu :)

eduramiba:
Sorry, the build is failing so I will upload this manually later.

eduramiba:
I uploaded this manually, can you please try to install it in Gephi? Thanks again!

andypiper:
Hi there, just to clarify, this plugin is using Twitter4J so I believe that the sample stream will be this one, rather than the newer one you referred to (the data format is also different on the newer sample stream, so you would need additional logic I suspect). https://developer.twitter.com/en/docs/tweets/sample-realtime/overview/get_statuses_sample

PR #225, opened 2021-05-11
1.4.3 : Bug fix for Indonesian language filter

totetmatt:
Official ISO 639-1 code for language define Indonesian as `id`.

But twitter is using old definition `in` (checked on received tweet)

ref : https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes

bug report : https://twitter.com/bikejoon/status/1391999609442996226


eduramiba:
I think I will need to do manual deployment though

eduramiba:
Hi Matthieu,

Sorry for the long wait as I forgot about the manual deployment... This should now be deployed and update available through Gephi desktop.

Can you check it please?

Thanks!!

PR #267, opened 2022-09-01
Orderedlayout

WouterSpekkink:

* [X] New Plugin (but it is also an updated version of an existing plugin; I wanted a different name).
* [ ] Update

The plugin creates layouts for graphs that have some meaningful order (recorded as a node variable). For example, if nodes have a chronological order, this layout can be used to visualize that chronological order. Indeed, it can also be used to represent other kinds of orders. For example, I am working on a plugin that calculates trophic levels of nodes, which can be treated as another kind of order variable.

There is an existing plugin that I authored, which is called the event graph layout. I would like to replace that plugin with the ordered layout, because its name doesn't cover its possible applications very well. Also, I added some features to this new version of the plugin, such as being able to invert the layout or change the direction of the layout (between horizontal and vertical). 
  
1. Import a dataset in which the nodes have a meaningful order defined as a numeric variable.
2. It would be best to if the dataset also has edges.
3. Tweak the layout options if necessary and then run the plugin.

* [ ] Did you write unit tests to test your plugin functionalities? -> Not sure what these are.


mbastian:
Thanks @WouterSpekkink this looks pretty good to me. When I tried it out I was a bit puzzled by the behavior of the layout algorithm. When I enable/disable some properties like vertical or strong gravity it doesn't always change the layout the way I expect. For instance, if you turn on vertical and then turn it off again it doesn't change anything. Maybe I'm not understanding what you're trying to do.

In the event layout plugin you also had a README. Do you think you could provide that as well? Thanks!

WouterSpekkink:
Hi @mbastian,
Thanks for having a detailed look. I see what you mean with the vertical layout toggle. That only makes a difference when you also turn on the "Set Force on free axis" option. Basically, the vertical layout toggle let's you switch between having the ordered layout on the x-axis or the y-axis. I used to just lay out everything from left to right, but I found that some people prefer to have a layout ordered from top to bottom.

The strong gravity also only works when forces are acting on the "free axis". Perhaps it is worth considering greying out those options when the force on the free axis is deactivated (is that possible)? It might prevent some confusion, although I think the confusion will quickly pass if you are purposefully using the plugin and got the hang of it.

I'll add the README. Probably tonight.

Best,

Wouter

mbastian:
@WouterSpekkink Thanks for the details. Would be good to reflect this on some documentation I guess :) 

Let me know when you added the README and I can then merge

mbastian:
@WouterSpekkink Any update? Cheers

WouterSpekkink:
Sorry for the big delay. There were just so many things (personal and professional) that asked for my attention.

mbastian:
Thanks a lot, the plugin is now published!

https://gephi.org/plugins/#/plugin/orderedlayout

WouterSpekkink:
Thanks! Also for fixing various shortcomings.
-- 
 Sent with Tutanota, enjoy secure & ad-free emails. 



28 Jan 2023, 13:34 by ***@***.***:

Issue #262, opened 2022-08-28
Publish to the Web plugin


mbastian:
The existing Sigma exporter is maintained by @computermacgyver. You can see the branches on [their fork](https://github.com/oxfordinternetinstitute/gephi-plugins). We need to decide first if we want to update this plugin or to write a new one. If we decide to extend the existing one, I've already created an up-to-date branch on this repository at `sigmaexporter-plugin`.

jacomyal:
At OuestWare, many people have asked us a solution to share networks online, using sigma v2. Our solution is [Retina](https://ouestware.gitlab.io/retina/beta/) ([source page](https://gitlab.com/ouestware/retina)). Its current functional scope is to allow sharing online a GEXF graph, but with custom filtering, coloring and sizing.

It is still in beta version, but ultimately:
- It reads any graph available online, as long as there is no CORS restriction (and [Gist](https://gist.github.com/) applies)
- The state (sizes, colors, filters...) are stored in the URL
- All Retina versions will coexist so that permalinks are persistent

One solution to provide a solution to share GEXF online using sigma v2 would be to have a wizard in Gephi that uploads the GEXF graph somewhere online (maybe on Gist, if we can allow users to login from the wizard itself), and then give them a URL to use.

We could also make something similar (exporting a .zip that contains Retina code and the graph file), but I'm afraid it will be harder to maintain.

jacomyal:
Here is a list of things that the plugin must do, if we go that way:

### 1. We must register Gephi (or this plugin) as a GitHub app, to enable saving Gists

The doc is [here](https://docs.github.com/en/developers/apps/building-github-apps/creating-a-github-app). This is not the biggest topic, but we must store the registered app ID somewhere.

Also, to prevent having to deal with displaying a web page inside Gephi, the **Enable Device Flow** option must be checked.

### 2. The plugin should allow user to log into GitHub Gist

This will allow uploading the GEXF graph file onto Gist, to share it online with Retina.
The process to authenticate a user (ie. retrieving a usable API token) is the following (as described [here](https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#device-flow)):

1. Gephi will ask for a user verification code to GitHub, and display to the user the code and a URL. The API call contains the app ID and the required scope (`gist`, in our case). More information [here](https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps).
[...]

PR #268, opened 2022-09-01
Web publish plugin

seinecle:

* [X] New Plugin
* [ ] Update

Enable the user to publish the network they are working on, to the web. See this issue for details of the implementation:
https://github.com/gephi/gephi-plugins/issues/262#issuecomment-1231627948
  
1. Go to File -> Export -> Publish to the Web
2. Follow the instructions in the "Setup" tab
3. Switch to the "Publish" tab and publish.


mbastian:
Nice @seinecle and sorry for the delay. I started looking into it and have some general comments and questions:

1. You currently don't configure the GEXF exporter. Are all the default settings what you would expect? For instance I don't think the viewer support dynamic graphs so we should probably turn this off.
2. How do you deal with [Gist's limits](https://docs.github.com/en/rest/gists/gists#truncation)? Seems like the limit could be reach fairly quickly with large graphs so we should probably deal with it. At least inform the user. Maybe we should also thing about strategies to reduce the file size? Maybe Sigma.js supports zipped's GEXF?
3. You have a number of hardcoded strings in your code like URLs, clientId etc. Could you turn all of those into constants for code clarity? (private static final String)

seinecle:
hi  @mbastian,

1. I didn't know the exporter can be configured. I feel like the export to gist should preserve the network exactly as the user is working on it in Gephi, so I would not restrain / limit the gexf to some features. That's rather up to Retina / Sigma to handle it in the way they need?
2. I was not aware of these limits. I'll have a look and yes, I will add warnings and will add ways to handle the exceptions related to limits
3. Good point, will do.

seinecle:
Did some verifications:

Gist has a [push limit at 100Mb per file](https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github) (when not using GLFS).

I used the plugin to publish a network which corresponds to a gexf file of 61Mb with 2000 nodes and 279,000 edges.

- the upload on Gist works fine, as expected
- Retina can't display it: no error msg, does the page doesn't load

During the Gephi week I had asked @jacomyal about rough upper limits for the size of a network to be displayed by Retina. It would be around 10k nodes and 20k edges - above this, the browser might get slower on some computers.

So it seems that the first limit is the network size that Retina can handle, not the max file size Gist accepts.

I propose that we set a limit to the size of the network to be published? Like: 20k nodes max, 40k edges max? If any of these limits is reached, the "Publish" button remains greyed out with an explainer.

Your comments are welcome @mbastian @jacomyal

mbastian:
@seinecle Thanks for verifying! It's a good news although it does contradict the Gist documentation, which I find strange. The link you posted is for GitHub and may not apply to Gist. 

Regarding the size limit I'm not a fan of forbidding the upload but maybe rather inform the user that the experience is not gonna be optimal. The node/edge counts is also only partially a factor in the file size. The number of attributes and their content can also be a bit factor.

seinecle:
Ok, so no hard limit but a warning then - I'll update the plugin today.

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/GephiPluginDesktopLogic.java
```diff
@@ -0,0 +1,49 @@
...
+import org.openide.NotifyDescriptor;
+import org.openide.util.NbBundle;
+import org.openide.util.lookup.ServiceProvider;
+
+/*
+
+ * @author Clement Levallois
+ */
+@ServiceProvider(service = ExporterClassUI.class)
+public final class GephiPluginDesktopLogic implements ExporterClassUI {
```
mbastian, line 22: Poor class name. Something more straightforward like `WebPublishExporterUI` would be clearer.

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/GephiPluginDesktopLogic.java
```diff
@@ -0,0 +1,49 @@
...
+import org.openide.DialogDisplayer;
+import org.openide.NotifyDescriptor;
+import org.openide.util.NbBundle;
+import org.openide.util.lookup.ServiceProvider;
+
+/*
+
+ * @author Clement Levallois
+ */
+@ServiceProvider(service = ExporterClassUI.class)
```
mbastian, line 21: You may want to add a position to make sure it always appear third in the menu, after the default options. position=200 or something like this.

modules/WebPublishPlugin/src/main/resources/net/clementlevallois/web/publish/plugin/controller/Bundle.properties
```diff
@@ -0,0 +1,47 @@
...
+JPanelWebExport.tabGithub.TabConstraints.tabTitle=Set up (to do just once)
+general.message.error.gexf_not_retrieved=Unable to create gexf file from your network.
+general.message.success_reset=The reset was successful.
+general.message.error.gist_creation=Error creating online gexf file. Error code\:
+OpenIDE-Module-Short-Description=This plugin allows you to publish, visualize and share your network on the web, directly from Gephi.
+JPanelWebExport.step6.title=4. Paste the code on this website\:
+general.message.error.empty_network=Your network is empty. Create a network to publish it.
+JPanelWebExport.step2.title=2. Login to your Github Account
+JPanelWebExport.step1.title=1. You need a Github account and be logged in
+general.message.plugin.name=Publish to the web
```
mbastian, line 38: As a convention, all actions menu that open a new menu/window should end with three dots "..."

modules/WebPublishPlugin/pom.xml
```diff
@@ -0,0 +1,127 @@
...
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.netbeans.utilities</groupId>
+                <artifactId>nbm-maven-plugin</artifactId>
+                <configuration>
+                    <licenseName>Apache 2.0</licenseName>
```
mbastian, line 101: licenseFile is missing

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/GlobalConfigParams.java
```diff
@@ -0,0 +1,23 @@
+/*
```
mbastian, line 1: Delete those comments

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/JPanelWebExport.java
```diff
@@ -0,0 +1,612 @@
...
+
+/**
+ *
+ * @author LEVALLOIS
+ *
+ * The logic of this plugin follows the logic laid out here:
+ * https://github.com/gephi/gephi-plugins/issues/262#issuecomment-1231627948
+ *
+ */
+public class JPanelWebExport extends javax.swing.JPanel {
```
mbastian, line 38: I wouldn't start a class name with JPanel. Rather start with the class purpose and finish by panel or JPanel.

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/JPanelWebExport.java
```diff
@@ -0,0 +1,612 @@
...
+
+    private JsonObject responseGithubConnectAction;
+    private JsonObject responseGithubUserCodeInput;
+    private String accessToken;
+    private String deviceCode;
+
+    private static final ResourceBundle bundle = NbBundle.getBundle(GephiPluginDesktopLogic.class);
+
+    public static final String COLOR_SUCCESS = "#45ba48";
```
mbastian, line 47: You have quite a bit of colors hardcoded here, and then Color.WHITE. Not a big deal but with dark mode this may cause issues

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/JPanelWebExport.java
```diff
@@ -0,0 +1,612 @@
...
+                        + "grant_type=urn:ietf:params:oauth:grant-type:device_code";
+                HttpRequest request = HttpRequest.newBuilder()
+                        .uri(URI.create(url))
+                        .header("accept", "application/json")
+                        .POST(HttpRequest.BodyPublishers.ofString(inputParams))
+                        .build();
+
+                boolean success = false;
+                long startTime = System.currentTimeMillis();
+                long maxDuration = 900_000;
```
mbastian, line 89: Is that 15 minutes? that sounds like a really large timeout

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/JPanelWebExport.java
```diff
@@ -0,0 +1,612 @@
...
+                    }
+                    currDuration = (float) (System.currentTimeMillis() - startTime) / (float) 1000;
+                    // the min duration recommended by Github between two polls is 5 seconds
+                    // -> https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#device-flow
+                    Thread.sleep(5200);
+                    publish(++loops);
+                }
+            } catch (IOException | InterruptedException ex) {
+                Exceptions.printStackTrace(ex);
+                jTextFieldGithubErrorMsg.setText(ex.getMessage());
```
mbastian, line 107: Within the `SwingWorker` class you're running on a separate thread than the EDT (the swing thread). All the Swing code should normally be sent back to the EDT, usually with `SwingUtilities.invokeLater()`. It will work but not fully correct.

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/JPanelWebExport.java
```diff
@@ -0,0 +1,612 @@
...
+        jTextAreaUrls.setText("");
+    }
+
+    SwingWorker pollWorker = new SwingWorker<Void, Integer>() {
+
+        @Override
+        protected Void doInBackground() throws Exception {
+            JsonObject jsonObject = new JsonObject();
+            try {
+                HttpClient client = HttpClient.newHttpClient();
```
mbastian, line 71: Did you try what this does when one is not connected to the internet?

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/PublishingActions.java
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * @author LEVALLOIS
+ */
+public class PublishingActions {
+
+    private static final ResourceBundle bundle = NbBundle.getBundle(GephiPluginDesktopLogic.class);
+
+    private static final int BYTES_IN_A_MEGABYTE = 1_048_576;
+
+    public static JsonObject getGexfAsString() {
```
mbastian, line 41: I don't get why you're using a `JsonObject` to pass data between this class and the client. Seems like your goal was to be able to pass the gexf data plus some kind of error codes with it. 
- The GEXF string seems to be converted back and forth multi times between JSON and String as a result
- Instead of error codes, a more standard way in Java would be to create your own exception

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/PublishingActions.java
```diff
@@ -0,0 +1,147 @@
...
+        int sizeGexfInBytes = gexfToSendAsString.getBytes(StandardCharsets.UTF_8).length; //
+
+        // Files pushed to a github shouldn't be larger than 100Mb
+        // There is still a doubt: is this limit the same for uploads to a gist?
+        if (sizeGexfInBytes >= (BYTES_IN_A_MEGABYTE * MAX_MB_FOR_GITHUB_PUSH)) {
+            jsonObject.addProperty(ERROR_CODE_GEXF_TOO_BIG, bundle.getString("general.message.error.network_too_big"));
+            return jsonObject;
+        }
+
+        if (graph.getNodeCount() > 10_000 || graph.getEdgeCount() > 20_000) {
```
mbastian, line 80: Those limits are hardcoded, not ideal IMO

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/PublishingActions.java
```diff
@@ -0,0 +1,147 @@
...
+        JsonObject fileItself = new JsonObject();
+        JsonObject contentsFile = new JsonObject();
+        contentsFile.addProperty("content", gexfFile);
+        fileItself.add(fileName, contentsFile);
+        bodyPostGexfToGist.add("files", fileItself);
+        String bodyToString = bodyPostGexfToGist.toString();
+
+        try {
+            HttpClient client = HttpClient.newHttpClient();
+            String url = "https://api.github.com/gists";
```
mbastian, line 124: Hardcoded URL

mbastian (changes requested):
I added some comments, mostly related to conventions. One general comment about the architecture is trying to better decouple the business logic from the UI. You have part of it in the `PublishingActions`, which I think is great as it allows testability. You can check the 3 unit tests I created for reference. On the other hand, some other logic is buried somewhere in the `JPanelWebExport` class, making it difficult to unit test. Ideally all logic is wrapped into clear functions that can be unit tested.

seinecle:
Wow indeed many improvements to make. My comments:

- on the json object that I use to pass stuff around. It is indeed weird and not ideal. You suggest that I could throw exceptions instead. Ok, good! I have never worked this way though, so I should get back to some Java core textbook. Will take a couple of days.
- on the SwingWorker / EDT stuff: will go and try what you suggest. I am not familiar at all with these things so any code example / resource would be great, by any chance.

modules/WebPublishPlugin/pom.xml
seinecle: ✅

modules/WebPublishPlugin/src/main/resources/net/clementlevallois/web/publish/plugin/controller/Bundle.properties
seinecle: ✅

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/GephiPluginDesktopLogic.java
seinecle: ✅

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/GephiPluginDesktopLogic.java
seinecle: ✅

modules/WebPublishPlugin/src/main/java/net/clementlevallois/web/publish/plugin/controller/PublishingActions.java
seinecle: If the URL ever changes, the query parameters might change as well. For this reason, I would rather keep in the same view / location both the URL and the query params.

mbastian:
@seinecle I did a deeper refactoring to help decouple the UI from the business logic, which was still tightly coupled. In addition, I uses our LongTask API instead of SwingWorker. With that, you easily get the progress bar too. The two `Runnable` classes are now completely decoupled from the UI. The UI updates itself through a `LongTaskListener` (success) and `LongTaskErrorHandler` (error).

One UI improvement I think you could still make is on the final result. It's really paintful to have to copy-paste the final URL from a text field. Why not having an "Open" button or "Copy to clipboard"? Java can easily open the browser for you with an URL.

seinecle:
@mbastian thanks! `LongTask `seems super nice, glad to discover this utility! And thanks for the refactoring.

I have:

- added the buttons you mention (open in web browser + copy link to clipboard).
- this needed to change slightly the way the `LongTask` was returning strings
- localized all strings
- renamed a class to WebExport JPanel

mbastian:
Thanks @seinecle! I just noticed that the access token we get from GitHub expires after 8 hours. This isn't something we can change but there is a [flow we can use via refresh tokens](https://docs.github.com/en/developers/apps/building-github-apps/refreshing-user-to-server-access-tokens) to have 6 months tokens. It seems unpractical to have to go through the authentification every day so it would be good to implement that. We shall also change the narrative in the UI. It's a little more complicated than "Setup once".

seinecle:
ok, thanks for identifying this! Will have a look.

seinecle:
@mbastian reading the doc, it seems we can just opt out of expiring tokens? If I understand correctly, the 8 hours limit is just for expiring tokens.
So in Gephi's github app ("gephi-lite"), there just needs to opt out expiring tokens. The steps to find this option are detailed in the link you included in your comment. Please let me know if I misunderstand something.

mbastian:
@seinecle Interesting. It still don't really get from the doc what happens when we opt-out? Access token don't expire at all? I opted out now. Let's see how it goes

seinecle:
@mbastian do you think we can move to the publication?

mbastian:
@seinecle Yes, almost :) Looks like you inverted the buttons between gist view versus network viz.

mbastian (approved):

mbastian:
@seinecle The plugin is published https://gephi.org/plugins/#/plugin/web-publish-plugin

Thanks for this great work! cc @jacomyal

seinecle:
@mbastian  Thank you for the extensive code review !!

DeJeroenBakker:
First of all, thanks for developing this plugin @seinecle. I can imagine this being very useful! 

The plugin doesn't seem to work for me at the moment, however. The authentication process works flawlessly, but when I try to publish my network, I always get the following 500 error:

`Error creating online gexf file. Error code:500; Error message:` 

Is there any way to fix this? I've tried re-authenticating, using a different network and restarting Gephi, but it doesn't seem to make a difference.

Thanks!

PR #273, opened 2022-09-02
Spatial Ranking comeback

mbastian:

* [ ] New Plugin
* [x] Update

Map attributes to X/Y/Z coordinates to create scatter plots.

Requested in #271 by @grandjeanmartin.

1. Load any graph, preferably with some columns
2. Go to Appearance and notice the extra icon for coordinates
3. Select Ranking and the column you want to map to coordinates
4. Select the axis and the min/max size
5. Press Apply

Note that I also added a "Simple" transformer that does the same thing as the random layout.

* [X] Did you write unit tests to test your plugin functionalities? -> Yes added 2 basic tests


mbastian:
@grandjeanmartin How would you suggest to rename this plugin? ;)

grandjeanmartin:
@mbastian you rock! For the name, feel free to use the one that suits you, to me something like « network to scatter plot » would be the most descriptive.

seinecle:
Indeed, I came here thinking of a plugin that would have to do with longitudes and latitudes, because of "spatial". Dropping " spatial" in the name could avoid that.

mbastian:
Ok, I go with `Position Ranking` then as a good comprimise.

PR #282, opened 2022-10-25
Multilayer Network Visualization.

JSiv:
New Plugin (MultiViz)

Purpose :Visualization of Multi-Layer Networks
  
## How to test the plugin in Gephi?

After Selecting the Plugin:
1. Select a Layer ( Features of the network)
2. Select a Layout Algorithm
3. Set Distance between two layers.

If Layout Algorithm Selected is a Force Directed Algorithm
1. Set Iterations, Speed, Gravity and Area(optional)

Split By Level
1.  True: Run the selected algorithm on whole network (only once before splitting the layer).
2.  False: Run the selected algorithm on each Layer.

More Details on   https://arxiv.org/pdf/2209.03149.pdf


mbastian (commented):
Hi @JSiv and thanks for this new plugin, looks exciting. 

My first comment is why did you copy all the existing layouts in your code? Why not just reuse the classes from the `LayoutPlugin` module?

JSiv:
> Hi @JSiv and thanks for this new plugin, looks exciting.
> 
> My first comment is why did you copy all the existing layouts in your code? Why not just reuse the classes from the `LayoutPlugin` module?


Hi @mbastian ,

Thank you.

The layouts here is a slightly modified version of the original codes. For force directed layouts, I tried to modify the `GraphModel API` to pass only a selected no of nodes to the algorithm, but it didn't work, so i copied the existing layout code and modified it based on my needs. Also, I have added the files as separate (resulting in some redundant codes) in order to add the appropriate copyright licenses.

mbastian:
> The layouts here is a slightly modified version of the original codes. For force directed layouts, I tried to modify the `GraphModel API` to pass only a selected no of nodes to the algorithm, but it didn't work, so i copied the existing layout code and modified it based on my needs. Also, I have added the files as separate (resulting in some redundant codes) in order to add the appropriate copyright licenses.

Thanks for the clarification. I'm not a fan of having this amount of code duplication. A bit of duplication is fine but here it's really a lot, even for a plugin. There must be another, more elegant way to achieve what you want. Can you be more specific on what you were trying to do but couldn't? Seems like running a layout on a subset of nodes was your intention?

JSiv:
> > Thanks for the clarification. I'm not a fan of having this amount of code duplication. A bit of duplication is fine but here it's really a lot, even for a plugin. There must be another, more elegant way to achieve what you want. Can you be more specific on what you were trying to do but couldn't? Seems like running a layout on a subset of nodes was your intention?


Hi @mbastian,

Yes, I was trying to pass a subset of nodes to the layout algorithm. Is there any way to do it?

JSiv:
Hi @mbastian


I've used `NodePartitionFilter` from `org.gephi.filters.plugin` to pass a subset of nodes to the layout algorithms and removed all the redundant codes. 

When the user want to create a visualization based on the any edge properties (eg: label, weight), a new column is added to the node table named `mviz_edge_*edge property*`. Using this newly created table  `NodePartitionFilter` is used to get the subset of nodes (nodes which are connected by edges with same property). And if a node is connected by different edges with different properties, the node is only shown in the first layer (i.e only first property is taken into consideration). .

PR #290, opened 2023-02-11
Seadragon plugin

totetmatt:

* [ ] New Plugin
* [X] Update

Able to see large map with opensea dragon.
  
1. Open Gephi, get a Spatialized Graph
2. File >> Export >> Seadragon Web ...
3. Export
4. Check that file are present in the exported directory
5. Use of webserver (``python -m http.server` is enough) to make sure it works

* [X] Did you write unit tests to test your plugin functionalities?


modules/SeaDragonPlugin/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>gephi-plugin-parent</artifactId>
+        <groupId>org.gephi</groupId>
+        <version>0.10.0</version>
+    </parent>
+
+    <groupId>org.gephi</groupId>
+    <artifactId>sea-dragon-plugin</artifactId>
```
mbastian, line 11: Not so important but if we want to match the previous plugin identified it should be "seadragon-web-export"

modules/SeaDragonPlugin/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+    <version>2.0.0</version>
+    <packaging>nbm</packaging>
+
+    <name>Sea Dragon Plugin</name>
```
mbastian, line 15: Probably "Seadragon Web Export" is clearer.

modules/SeaDragonPlugin/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.netbeans.utilities</groupId>
+                <artifactId>nbm-maven-plugin</artifactId>
+                <configuration>
+                    <licenseName>Apache 2.0</licenseName>
+                    <author>mbastian</author>
```
mbastian, line 96: Add yourself? :)

modules/SeaDragonPlugin/pom.xml
```diff
@@ -0,0 +1,121 @@
...
+                    <authorEmail></authorEmail>
+                    <authorUrl></authorUrl>
+                    <sourceCodeUrl>https://github.com/totetmatt/gephi-plugins.git</sourceCodeUrl>
```
mbastian, line 99: Should be "https://github.com/gephi/gephi-plugins/tree/seadragon-plugin"

mbastian (commented):
Great work! Two comments
- Unify the naming. Sometimes it's "SeaDragon", sometimes it's "Seadragon" and sometimes "sea-dragon".
- I guess the resources/org/gephi/plugins/seadragon folder can be deleted?

totetmatt:
> Great work! Two comments
> 
>     * Unify the naming. Sometimes it's "SeaDragon", sometimes it's "Seadragon" and sometimes "sea-dragon".
> 
>     * I guess the resources/org/gephi/plugins/seadragon folder can be deleted?

About naming, If I understand seadragon was the original name, but it got bought by Microsoft, and now the tool available is called opeseadragon. 
Should we adapt with the "openseadragon" name or keep the old one for "inheritance" purpose ?

mbastian:
> About naming, If I understand seadragon was the original name, but it got bought by Microsoft, and now the tool available is called opeseadragon. Should we adapt with the "openseadragon" name or keep the old one for "inheritance" purpose ?
> 
> Also maybe "export" would be better as "web export" implies the plugin also do web publication, which would avoid confusion ?

Yes, maybe better to use "OpenSeadragon" instead. The legacy plugin isn't available since 0.8 so I think inheritance is not so important. So "OpenSeadragon Export" seems good to me.

totetmatt:
https://github.com/gephi/gephi-plugins/pull/291 renaming everything so now it's here

PR #294, opened 2023-03-22
word cloud plugin for review and publication

seinecle:

* [ X] New Plugin
* [ ] Update

Showing a word cloud of a textual node attribute, while the user is hovering over the network with their mouse.

  
1. Open a network with a textual attribute. A test network is provided here.
2. Click on "refresh" in the plugin to get a list of the attributes. For the test, choose "description".
3. Click on start and follow the instructions

* [X ] Did you merged with the `master` branch to get the latest updates?
* [ ]X Did you build and test the plugin successfully locally?
* [ X] Did you include metadata (e.g. plugin author, license) in the `pom.xml` file
* [ X] Did you make sure NOT to include any unecessary files in your PR?
No, the reason is that I tested it extensively, also with the help o f Dr. Veronica Espinoza, and many checks for null exceptions were added.


pom.xml
```diff
@@ -7,11 +7,12 @@
     <version>0.10.0</version>
     <packaging>pom</packaging>
 
-    <name>gephi-plugins</name>
+    <name>parent project for word cloud plugin</name>
```
mbastian: What is that?

modules/WordCloudPlugin/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+        <gephi.version>0.10.0</gephi.version>
+        <netbeans.version>RELEASE160</netbeans.version>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+    </properties>
+
+
+    <dependencies>
+        <dependency>
+            <groupId>net.clementlevallois.functions</groupId>
```
mbastian, line 18: Where are those `umigon` dependencies coming from? If they are not on Maven central they won't be found when building.

modules/WordCloudPlugin/pom.xml
```diff
@@ -0,0 +1,124 @@
...
+        <version>0.10.0</version>
+    </parent>
+
+    <groupId>exploreyourdata</groupId>
+    <artifactId>word-cloud-plugin</artifactId>
+    <version>0.10.0</version>
+    <packaging>nbm</packaging>
+    <name>Word Cloud Plugin</name>
+    
+    <properties>
```
mbastian: Seems like this properties are duplicated from the parent POM. Delete them and it should still work.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/DataManager.java
```diff
@@ -0,0 +1,27 @@
+/*
```
mbastian, line 1: Would be nice to cleanup license headers

modules/WordCloudPlugin/pom.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>gephi-plugin-parent</artifactId>
+        <groupId>org.gephi</groupId>
+        <version>0.10.0</version>
+    </parent>
+
+    <groupId>exploreyourdata</groupId>
```
mbastian: For some other plugins you used `net.clementlevallois` as groupId. Would be great to be consistent between all of your plugins.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/GraphOperations.java
```diff
@@ -0,0 +1,41 @@
...
+        if (pc.getCurrentProject() == null) {
+            return null;
+        }
+        Workspace workspace = pc.getCurrentWorkspace();
+        GraphModel gm = Lookup.getDefault().lookup(GraphController.class).getGraphModel(workspace);
+        return gm;
+    }
+
+    public static DefaultListModel<String> returnTextualNodeAttributesAsListOfNames(GraphModel gm) {
+        List<Column> listOfNodeAttributes = gm.getNodeTable().toList();
```
mbastian: Quite verbose. Using stream filter would be a bit nicer

```
gm.getNodeTable().toList().stream().filter((column) -> (column.getTypeClass() == String.class)).forEach((column) -> {
        listModel.addElement(column.getId());
});
```

Also, technically usint `equals` is more correct `column.getTypeClass().equals(String.class)`

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/InitialWordProcessingRunnable.java
```diff
@@ -0,0 +1,50 @@
...
+import org.gephi.graph.api.GraphModel;
+import org.gephi.utils.longtask.spi.LongTask;
+import org.gephi.utils.progress.Progress;
+import org.gephi.utils.progress.ProgressTicket;
+import org.openide.util.NbBundle;
+
+public class InitialWordProcessingRunnable implements LongTask, Runnable {
+
+    private ProgressTicket progressTicket;
+    private boolean cancelled = false;
```
mbastian: This field is not used, should be deleted

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/StaticProperties.java
```diff
@@ -0,0 +1,27 @@
...
+ * Click nbfs://nbhost/SystemFileSystem/Templates/Licenses/license-default.txt to change this license
+ * Click nbfs://nbhost/SystemFileSystem/Templates/Classes/Class.java to edit this template
+ */
+package net.clementlevallois.wordcloud;
+
+/**
+ *
+ * @author LEVALLOIS
+ */
+public class StaticProperties {
```
mbastian, line 7: Any reasons those properties aren't all final?

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopWordsFinderRunnable.java
```diff
@@ -0,0 +1,110 @@
...
+    public TopWordsFinderRunnable(Integer pauseBetweenComputations, int topWordsToRetrieve) {
+        this.pauseBetweenComputations = pauseBetweenComputations;
+        this.topWordsToRetrieve = topWordsToRetrieve;
+    }
+
+    @Override
+    public void run() {
+        Progress.start(progressTicket);
+        Progress.setDisplayName(progressTicket, bundle.getString("expression.warning.wordcloud_analysis_running"));
+        while (true && !this.cancelled) {
```
mbastian: true can be removed

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopWordsFinderRunnable.java
```diff
@@ -0,0 +1,110 @@
...
+                        Exceptions.printStackTrace(ex);
+                    }
+                    continue;
+                }
+                // retrieve the textual attributes of these nodes
+                // and compute the top terms of these
+                TopTermExtractor topTermExtractor = new TopTermExtractor();
+                List<String> selectedNodesIds = new ArrayList(setIdsForTestChange);
+                String topTermsExtractorFromSelectedNodes = topTermExtractor.topTermsExtractorFromSelectedNodes(selectedNodesIds, topWordsToRetrieve);
+                try {
```
mbastian: Isn't there a way to refactor this method so the `Thread.sleep` isn't repeated 3 times?

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopWordsFinderRunnable.java
```diff
@@ -0,0 +1,110 @@
...
+import org.gephi.utils.progress.Progress;
+import org.gephi.utils.progress.ProgressTicket;
+import org.gephi.visualization.VizController;
+import org.openide.util.Exceptions;
+import org.openide.util.NbBundle;
+
+public class TopWordsFinderRunnable implements LongTask, Runnable {
+
+    private ProgressTicket progressTicket;
+    private final LinkedBlockingQueue<String> results = new LinkedBlockingQueue();
```
mbastian, line 23: In multiple places you are using the `node.id` as a key. Or passing a list of node ids. I think it would be easier to comprehend if you were using the Node object directly, instead of it's id.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopWordsFinderRunnable.java
```diff
@@ -0,0 +1,110 @@
...
+
+    public String getIntermediary() throws InterruptedException {
+        return results.take();
+    }
+
+    @Override
+    public boolean cancel() {
+        this.cancelled = true;
+        Progress.finish(progressTicket, bundle.getString("expression.warning.wordcloud_analysis_stopped"));
+        return false;
```
mbastian: Your task is cancellable as you use the `cancelled` variable to stop the task in `run()`. Therefore, you should return `true`.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopWordsFinderRunnable.java
```diff
@@ -0,0 +1,110 @@
...
+import org.openide.util.NbBundle;
+
+public class TopWordsFinderRunnable implements LongTask, Runnable {
+
+    private ProgressTicket progressTicket;
+    private final LinkedBlockingQueue<String> results = new LinkedBlockingQueue();
+    private final Integer pauseBetweenComputations;
+    private Integer topWordsToRetrieve = StaticProperties.DEFAULT_WORDS_TO_DISPLAY;
+    private boolean cancelled = false;
+    private Set<String> previousSelectedNodes = new HashSet();
```
mbastian: Should not be a class member as it's just used by `run()`

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopTermExtractor.java
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * @author LEVALLOIS
+ */
+public class TopTermExtractor {
+
+    public static void main(String args[]) throws IOException, Exception {
+        new TopTermExtractor().importFromFile();
+    }
+
+    public void importFromFile() throws IOException, Exception {
```
mbastian: I'm not understanding what is this about? Maybe something you used for testing and forgot to remove?

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopTermExtractor.java
```diff
@@ -0,0 +1,186 @@
...
+        // we will iterate / loop on the nodes of the graph
+        NodeIterable nodes = graph.getNodes();
+
+        // we will store the text of the attribute for each node in a list
+        Map<String, String> textsFromTheAttribute = new HashMap();
+
+
+        // doing the iteration now
+        Iterator<Node> iteratorOnNodes = nodes.iterator();
+        while (iteratorOnNodes.hasNext()) {
```
mbastian: Using a for loop is more standard here
```
for(Node node : nodes) {
}
```

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopTermExtractor.java
```diff
@@ -0,0 +1,186 @@
...
+                return false;
+            }
+        }
+        DataManager.setMapOfNodeIdsToTheirTextFragments(mapOfNodeIdsToTheirTextFragments);
+        return true;
+    }
+
+    public String topTermsExtractorFromSelectedNodes(List<String> nodeIds, int maxNumberOfTerms) {
+        // the multiset will store unique terms from the text we collected and count how many times each term appears
+        Multiset<String> allTextFragmentsFromAllSelectedNodes = new Multiset();
```
mbastian: I think there is a more elegant way to perform this whole method using streams. If I understand correctly, you construct a list of strings, then you sort it and want to retrieve the top list with its number of occurrences.

By prompting chat gpt with "Using Java Stream APIs and taking a list of strings as input, how do I obtain a sorted top-n list with occurrences count?" you should get a good answer.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/LexplorerTopComponent.java
```diff
@@ -0,0 +1,551 @@
...
+    private javax.swing.JLabel loggingjTextField;
+    private javax.swing.JInternalFrame numberTopWordsFrame;
+    private javax.swing.JPanel parametersPanel;
+    private javax.swing.JLabel placeHolderForTopTerms;
+    private javax.swing.JButton runButton;
+    private javax.swing.JTabbedPane tabbedPane;
+    private javax.swing.JPanel wordCloudPanel;
+    // End of variables declaration//GEN-END:variables
+    @Override
+    public void componentOpened() {
```
mbastian, line 462: Can be removed

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/LexplorerTopComponent.java
```diff
@@ -0,0 +1,551 @@
...
+        // TODO add custom code on component opening
+    }
+
+    @Override
+    public void componentClosed() {
```
mbastian, line 466: Can be removed

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/LexplorerTopComponent.java
```diff
@@ -0,0 +1,551 @@
...
+        // TODO add custom code on component closing
+    }
+
+    void writeProperties(java.util.Properties p) {
```
mbastian: Can be removed, and same for `readProperties`

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/LexplorerTopComponent.java
```diff
@@ -0,0 +1,551 @@
...
+                    uiUpdaterAsSwingWorker.execute();
+                    executor.execute(topWordsRetrieverAsRunnable, topWordsRetrieverAsRunnable);
+                    logAreaUpdater = new LogAreaUpdater(bundle.getString("expression.switch_to_wordcloud_tab_and_select"));
+                    logAreaUpdater.execute();
+                }
+            });
+            initialWordProcessingRunnable = new InitialWordProcessingRunnable(graphModel, selectedColumnId, StaticProperties.DEFAULT_TEXT_LANGUAGE);
+            executor.execute(initialWordProcessingRunnable, initialWordProcessingRunnable);
+        } else {
+            if (initialWordProcessingRunnable != null) {
```
mbastian, line 373: No need to call call on the Runnable if you already call `executor.cancel`

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/InitialWordProcessingRunnable.java
```diff
@@ -0,0 +1,50 @@
...
+        TopTermExtractor topTermExtractor = new TopTermExtractor();
+        topTermExtractor.tokenizeSelectedTextualAttributeForTheEntireGraph(graphModel, columnName, lang);
+        Progress.finish(progressTicket, bundle.getString("progress.initial_analysis_complete"));
+
+    }
+
+    @Override
+    public boolean cancel() {
+        this.cancelled = true;
+        Progress.finish(progressTicket, bundle.getString("progress.initial_analysis_interrupted"));
```
mbastian, line 43: In your current configuration, this runnable won't be interrupted. If you want the `LongTaskExecutor` to kill your Thread when `cancel` is called, you need to initialise it with a `interruptDelay`, for instance 
```new LongTaskExecutor(true, "Executor", 1000)```. In this example, the thread will be killed 1 second after cancel has been hit. If no interrupt delay is passed, the executor doesn't interrupt so cancel has essentially no effect.

That all said, it seems like a very bad idea to interrupt the thread (ie. a dirty cancel) as you are iterating over the nodes. While you iterate over the nodes you hold a read lock on the graph. If the thread is interrupted, the read lock won't be released and Gephi will eventually hang. 

I would advise to either not make this task cancellable and deal with it, or refactor so that the `tokenizeSelectedTextualAttributeForTheEntireGraph` method can be cancelled.

File: modules/WordCloudPlugin/pom.xml [hunk 22]
```diff
@@ -0,0 +1,124 @@
...
+    <artifactId>word-cloud-plugin</artifactId>
+    <version>0.10.0</version>
```
mbastian: I wouldn't recommend to use Gephi's versions for a plugin. Why not `1.0.0.`?

modules/WordCloudPlugin/README.md
```diff
@@ -0,0 +1,6 @@
+## Word Cloud Plugin
```
mbastian, line 1: Technically speaking, you are not showing a World Cloud at the moment. I guess that's still the intention for a future update? If not, maybe consider renaming the plugin

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/LexplorerTopComponent.java
```diff
@@ -0,0 +1,551 @@
...
+@ConvertAsProperties(
+        dtd = "-//net.clementlevallois.wordcloud//Lexplorer//EN",
+        autostore = false
+)
+@TopComponent.Description(
+        preferredID = "LexplorerTopComponent",
+        //iconBase="SET/PATH/TO/ICON/HERE",
+        persistenceType = TopComponent.PERSISTENCE_ALWAYS
+)
+@TopComponent.Registration(mode = "filtersmode", openAtStartup = true)
```
mbastian: I'm not sure it makes sense for plugins to open at startup. I don't have a strong opinion. On one hand it makes it easier to find. On the other hand if every plugin opens new components that ends up not being used during the session it can slow down Gephi.

mbastian (changes requested):
Congrats on getting to the finishing line submitting the plugin. I've added a few remarks, let me know if anything is unclear.

modules/WordCloudPlugin/pom.xml
seinecle: Actually they are added to the nbm, I think? I tested the nbm on  my gephi install and it works?
These are personal projects open sourced on Github, which are discovered at build time because they are in my Maven repo on my local machine.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/DataManager.java
seinecle: Ah yes I always forget this one

modules/WordCloudPlugin/pom.xml
seinecle: Absolutely, will revert to net.clementlevallois

modules/WordCloudPlugin/pom.xml
seinecle: Indeed

modules/WordCloudPlugin/README.md
seinecle: Ahhh... Well... This is both an intent (a future version might display words on two axes), and a way to make the function of the plugin immediately clear: summarizing an unstructured text by showing its most frequent words.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/LexplorerTopComponent.java
seinecle: I took a default template for the plugin, that is a default value. Ok I'll turn it to false!

pom.xml
seinecle: Probably a mistake :-)

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/StaticProperties.java
seinecle: No...

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopTermExtractor.java
seinecle: Indeed

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopTermExtractor.java
seinecle: I have a fondness for Iterators but indeed it is an overkill there

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopWordsFinderRunnable.java
seinecle: Ok I thought a simple id was saving on memory

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopWordsFinderRunnable.java
seinecle: Indeed

modules/WordCloudPlugin/pom.xml
mbastian: Yes, it works on your local machine but we need to build this plugin in GitHub so it can be released. Without these dependencies in a Maven repository we can't build it. Best would be if your libraries are published to Maven Central (see [guide](https://maven.apache.org/repository/guide-central-repository-upload.html)).

seinecle: Woot? Publishing to Maven central is a big step. Ok, will do.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopWordsFinderRunnable.java
mbastian: That won't make a difference so sounds like a premature optimisation. Code readability is generally more important than these small optimisations, which usually don't matter anyway.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopTermExtractor.java
mbastian: Fundamentally the underlying code is the same but for loops are considered more standard AFAIK.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/LexplorerTopComponent.java
mbastian: Yes, I believe the default is `true`. Maybe we should change that.

pom.xml
seinecle: that was a convenience for me to distinguish vetween several parent poms for Gephi plugins. Can't find where this is now.

modules/WordCloudPlugin/pom.xml
seinecle: done!

pom.xml
seinecle: ok it was to help me distinguish between several netbeans projects names in the same way

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/InitialWordProcessingRunnable.java
seinecle: I have added a Boolean flag in the while loop which iterates over the nodes. This flag is turned to 'false' when the user clicks on the cancel Button, before excecutor.cancel() is called.

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopTermExtractor.java
seinecle: I suggest to keep it (minus the file name), it helps me maintain it

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopWordsFinderRunnable.java
seinecle: done, also simplified the logic

modules/WordCloudPlugin/src/main/java/net/clementlevallois/wordcloud/TopTermExtractor.java
seinecle: ChatGPT was perfect for that. It removed the need for a Multiset object that I imported from a utils library just for that - so one less dependency for the plugin 💯 .

mbastian:
Thanks for addressing the comments. I added a few commits, check it out. I wanted to show you how to use Unit tests for your plugin, instead of creating `main` methods with hardcoded filenames. 

I added a `TopTermExtractorTest` file with a simple test but I was not able to make it work. It produces an exception, 
```
java.lang.NullPointerException
	at java.base/java.io.Reader.<init>(Reader.java:167)
	at java.base/java.io.InputStreamReader.<init>(InputStreamReader.java:109)
	at net.clementlevallois.umigon.tokenizer.controller.PatternOfInterestChecker.loadPatternsOfInterest(PatternOfInterestChecker.java:32)
	at net.clementlevallois.umigon.tokenizer.controller.UmigonTokenizer.tokenize(UmigonTokenizer.java:52)
	at net.clementlevallois.wordcloud.TopTermExtractor.tokenizeSelectedTextualAttributeForTheEntireGraph(TopTermExtractor.java:78)
	at net.clementlevallois.wordcloud.TopTermExtractorTest.testOnGraph(TopTermExtractorTest.java:16)
```

Not sure why this might only fail with the unit test while the plugin is working. In any event it's clear that the [patterns_of_interest.txt](https://github.com/seinecle/umigon-tokenizer/blob/main/src/main/java/net/clementlevallois/umigon/tokenizer/controller/patterns_of_interest.txt) file should not be located in the `src/main/java` folder but in the `src/main/resources`. That is what causes the exception I think.

Then, regarding design and testability. The fact you're using a static `DataManager` in this method is far from ideal. My advice when possible is to stick to method return types so it's really easy to test a method output.

seinecle:
Hello,
Merci pour le temps pris! Ok je ferai mes tests de cette facon, bien compris. J'ai mis a jour la dependence pour que les fichiers soient bien dans le repertoire ressources. Pour le DataManager, j'avoue que je suis un peu fatigue sur le refactoring la, mais est-ce que tu peux me dire quelle alternative tu as en tete : en faire une variable globale, non statique, dans le top component par exemple?

seinecle:
I have added a `readlock `call on the graph before the `unreadlock `method otherwise the test threw an exception

mbastian:
> Hello, Merci pour le temps pris! Ok je ferai mes tests de cette facon, bien compris. J'ai mis a jour la dependence pour que les fichiers soient bien dans le repertoire ressources. Pour le DataManager, j'avoue que je suis un peu fatigue sur le refactoring la, mais est-ce que tu peux me dire quelle alternative tu as en tete : en faire une variable globale, non statique, dans le top component par exemple?

No worries, it was more for future looking projects and learning how to do it. You can still add more tests if you want. This way, you'll also see by yourself that some patterns are easier to tests that others. No need to refactor the DataManager at this point. A good next step learning for you might be to dive in more in the concept of Model + Controller. Here, the DataManager stuff would be part of the Model.

> I have added a readlock call on the graph before the unreadlock method otherwise the test threw an exception

Thanks, I replaced the `readUnlock` with a `readUnlockAll`. This one doesn't throw an exception and is good when we're not sure if we're holding a lock or not. That is the case here. If the loop finishes normally we don't hold a lock (the NodeIterable takes care of releasing it for us). If the loop gets cancelled, we hold a lock.

mbastian (approved):

PR #297, opened 2023-03-30
Fix for 2 critical bugs

seinecle:

* [ ] New Plugin
* [ X] Update

Word cloud
  
## Bug fix

The error was noticed by Veronica Espinoza right after the public release of the plugin: no word cloud is displayed.

The reason for this error, despite the plugin working well at release time, is that I was testing it locally on a wrong (dated) version, instead of the latest version I was working on. Why? Because I was testing the plugin by doing a build on the parent pom, instead of a a **clean** and build.

After investigation, the plugin does not throw an erorr or crashes, it is just that a condition is always false and as a consequence the methods triggering the display of the word cloud are never run.

-> I have fixed this condition (a Boolean that was inverted)

Then, the wordcloud gets displayed but it always returns the same list of words, which are the top words for the entire graph - not just the nodes being selected. Again, that was due to me testing the plugin on a wrong build, I did not notice it.

Now I made sure I was running the proper version of the plugin, where the selection of top nodes (instead of all nodes) has been correctly implemented.

1. Install, open a network where nodes have a textual attribute. [Example of such a file](https://github.com/gephi/gephi-plugins/files/11110144/seinecle_hydrated.txt) - download it and replace the txt extension with a gexf extension

2. Select the "description" attribute and run the plugin

* [X ] Did you merged with the `master` branch to get the latest updates?
* [ X] Did you build and test the plugin successfully locally?
On this step: the mistake I made in the previous release is that I tested the plugin in NetBeans by running "build" on the parent pom, instead of "clean and build". It did not fetch the latest version of the plugin and for this reason I did not see that my latest code development had introduced bugs.

* [X ] Did you include metadata (e.g. plugin author, license) in the `pom.xml` file
* [ X] Did you make sure NOT to include any unecessary files in your PR?
* [X ] Did you write unit tests to test your plugin functionalities?
Yes I added a test which makes sure that the method fetching the text of selected nodes actually does that.


mbastian:
I merged but the PR wasn't properly configured. The target branch should be `master-forge` not `master`

seinecle:
Oops sorry

PR #332, opened 2026-06-30
New Version of BlueskyGephi

totetmatt:

* [] New Plugin
* [X] Update

- Making gephi 0.11.1 compatible
- Can connect to different host (like eurosky.social)
  
1.
2.

* [ ] Did you merged with the `master` branch to get the latest updates?
* [ ] Did you build and test the plugin successfully locally?
* [ ] Did you include metadata (e.g. plugin author, license) in the `pom.xml` file
* [ ] Did you make sure NOT to include any unnecessary files in your PR?


eduramiba (approved):
Looks good to me!

Unrelated: now it's necessary to compile with java 17 to avoid compilation issue in other plugin (error in `modules/Multimode-Networks/src/main/java/cz/cvut/fit/gephi/multimode/MultiModeWindowTopComponent.java`).

Also, we should update junit it seems, or build with skipTests, otherwise you will get:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.6.0-M1:test (default-test) on project streaming-impl: The used JUnit Version 4.10 is not supported anymore. Please update to version 4.12+
```

But I think code in this plugin to be merged is fine, and the build process of master-forge will also be fine because it uses java 17 and does not seem to run tests.

PR #337, opened 2026-08-28
Add Gephi AI plugin 1.3.0

MattArtzAnthro:

* [x] New Plugin
* [ ] Update

Gephi AI adds a local HTTP API to Gephi Desktop so an AI assistant can drive a
running Gephi: build and edit graphs, run layouts and statistics, style, filter,
export, and read the analyst's on-canvas selection. The assistant connects
through a separate Model Context Protocol server, distributed independently; this
plugin does not require it to build, install, or be reviewed.

The API listens on 127.0.0.1 only. It refuses any request whose `Host` header is
not a loopback address, which blocks DNS rebinding, and any request carrying
`Origin` or `Sec-Fetch-Site`, which blocks a page the user is merely visiting
from driving the API with a cross-origin `fetch`. Browsers set those headers and
page JavaScript cannot forge them; local clients send neither.

Beyond that the API is not authenticated: it trusts every local process, the same posture as
the Graph Streaming plugin's server. That is deliberate, because the client is a local MCP
server, and the module README states it plainly for users.

Companion server and documentation: https://github.com/MattArtzAnthro/gephi-ai

1. Build and launch: `mvn clean package`, then
   `mvn org.gephi:gephi-maven-plugin:run` from the repository root.
2. Open **Tools > Gephi AI Server**. The dialog shows the state (running or
   stopped) and the listening URL. The server starts automatically, so it should
   read `running` on `http://127.0.0.1:8080`.
3. Confirm the API answers: `curl http://127.0.0.1:8080/health`. Expect JSON with
   `"service": "Gephi AI API"` and `"status": "running"`.
4. Confirm a browser cannot drive it:
   `curl -i http://127.0.0.1:8080/graph/stats -H "Origin: https://example.com"`.
   Expect `403 Forbidden`. Without the header the same call succeeds.
5. Exercise it end to end without an assistant. Create a project, add two nodes
   and an edge, then read the graph back:
   ```
   curl -X POST http://127.0.0.1:8080/project/new -H 'Content-Type: application/json' -d '{"name":"demo"}'
   curl -X POST http://127.0.0.1:8080/graph/nodes/add -H 'Content-Type: application/json' -d '{"nodes":[{"id":"a","label":"A"},{"id":"b","label":"B"}]}'
   curl -X POST http://127.0.0.1:8080/graph/edges/add -H 'Content-Type: application/json' -d '{"edges":[{"source":"a","target":"b"}]}'
   curl http://127.0.0.1:8080/graph/stats
   ```
   The nodes and the edge appear in the Gephi window as they are added.
6. In the dialog, click **Stop**. `curl http://127.0.0.1:8080/health` should now
   fail to connect. Click **Start** to bring it back, then repeat step 5 to confirm the
   restarted server still serves requests. Changing the port in the dialog takes effect
   on the next start.

* [x] Did you merge with the `master` branch to get the latest updates?
  Branched from `master-forge` at 47b77d2 and rebuilt against it, so the diff is
  the module and one line of the root POM, with no unrelated files.
* [x] Did you build and test the plugin successfully locally?
  `mvn -B package` on the full reactor: BUILD SUCCESS, 58 modules, 0 failures.
  67 JUnit tests pass in this module.
* [x] Did you include metadata (author, license) in the `pom.xml`?
  Apache 2.0, with `licenseFile` so the text is shown at install, plus author,
  homepage, and source URL. Every source file carries the licence header.
* [x] Did you make sure NOT to include any unnecessary files in your PR?
  24 files: the module and one added `<module>` line. No deletions.
* [x] Did you write unit tests to test your plugin functionalities?
  67 tests, including the loopback and browser-origin guards, graph operations,
  lock behaviour under real two-thread contention, and a smoke test that boots
  the server on an ephemeral port and exercises it over HTTP.

## Notes for reviewers

**Architecture.** The plugin registers no Layout, Statistics, or Filter service, because there
is no extension point for a background service, so a reviewer scanning the diff will not find a
`@ServiceProvider`. The module declares an `OpenIDE-Module-Install` lifecycle hook instead,
which starts and stops the listener with the module and gives `closing()` for an orderly
shutdown, plus a Tools menu action so the user can see and control it. The Graph Streaming
plugin is this repository's precedent for an embedded HTTP server inside Gephi.

**Reflection.** Three field lookups in `GephiControlService` reach the
`ReentrantReadWriteLock` behind `GraphLock`. `GraphLock` exposes no timed
acquisition, and without a timeout a leaked read hold wedges the session. Each
lookup is cached, wrapped, and falls back to the public blocking API on any
failure, and a unit test fails loudly if graphstore renames the field rather than
letting the plugin degrade silently. The workaround is removed once
https://github.com/gephi/graphstore/pull/294 lands, which adds `tryReadLock` and
`tryWriteLock` to the public API. Reflection is also used to configure statistics and the PNG
exporter, whose implementation classes expose no configuration API, to pause the renderer, and
in a selection-read fallback for builds whose visualization model does not carry the selection.
All of it is wrapped, and every path degrades to a no-op or an honest error rather than
failing the request.

**Threading.** Requests are served on HTTP threads. Graph mutations and file import run
there, under the lock, because the graph model does not require the event dispatch thread.
Preview refresh, screenshots, window and camera operations, and the PDF and SVG export
pipelines hop to it.

Style and structure changes are welcome. Edits from maintainers are enabled, so
please push directly to the branch rather than routing changes through me.

### DIFF
--- a/modules/GephiAI/LICENSE
+++ b/modules/GephiAI/LICENSE
@@ -1,0 +1,190 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to the Licensor for inclusion in the Work by the copyright owner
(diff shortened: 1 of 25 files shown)


MattArtzAnthro:
@mbastian need anything from me on this?

